PR #44, opened 2026-04-24
fix: correct tool count (18→17) and add missing palinode_prompt tool

nullinject:
修复 docs/INSTALL-CLAUDE-CODE.md 中的工具数量错误：
- 描述从 '18 tools' 改为 '17 tools'（与 README 一致）
- 在工具列表末尾补充缺失的 palinode_prompt 条目

工具列表现在与 README.md 中的 17 个工具完全一致。

### DIFF
--- a/docs/INSTALL-CLAUDE-CODE.md
+++ b/docs/INSTALL-CLAUDE-CODE.md
@@ -1,6 +1,6 @@
 # Installing Palinode with Claude Code
 
-Palinode gives Claude Code persistent memory via MCP — 18 tools for searching, saving, and managing memories across sessions. The `palinode-session` skill auto-captures milestones and decisions during coding, so your memory stays fresh without manual effort.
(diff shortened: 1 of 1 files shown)


Paul-Kyle:
Thanks for the catch! Both changes here have already landed in a subsequent commit — palinode_prompt is already in the table, and the tool count has moved past 18. The current count is actually 25 (per mcp.py), and the doc intro now reads 21, so the doc itself still has a gap worth noting. Four tools are missing from the table entirely: palinode_cluster_neighbors, palinode_depends, palinode_timeline, and palinode_topic_coverage. Closing this PR as superseded — those undocumented tools are tracked separately.

Paul-Kyle:
Hey @nullinject — thanks for the catch on this one. Quick context for posterity now that this is closed:

- The `palinode_prompt` row is already present in the table on `main`
- The tool count moved past 18 — current count is **25**, doc intro currently reads 21, so the gap will be reconciled in a follow-up doc fix that updates the count and adds the four still-undocumented tools (`palinode_cluster_neighbors`, `palinode_depends`, `palinode_timeline`, `palinode_topic_coverage`)

Appreciate you flagging it.

PR #79, opened 2026-08-03
fix(claims): display partial hash marker in claims resolution output (#66)

### LINKED ISSUE #66: Claims renderer doesn't show when a verification was only partial
`resolve_memory_claims` now returns `span_partial: true` when a claim's span had no stored `quote_hash` (only drift was checkable), but `format_claims_resolution` — the text renderer the CLI and MCP surfaces share — doesn't display it.

A hash-less `ok` currently renders identically to a fully verified one, which is exactly the distinction the field exists to make.

## Task

Show a `(partial — no stored hash)` marker on affected claims in the rendered output, with a small test.

## Why this is a good first issue

This is one function with existing tests to copy from in `tests/test_claims_508.py`.

fprobstv:
Adds the `(partial — no stored hash)` marker to claims where `span_partial=True` in `format_claims_resolution`. Previously, hash-less claims rendered identically to fully verified ones.

Closes #66

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -115,6 +115,7 @@ reader before writing new records with v0.10.0.
   warning naming the file, the offending value, and the accepted set, reports
   `layer_hint_ignored` in the `split_file` result, and counts `hints_ignored` in the
   sweep's stats — so a run that fell back to heuristics no longer reports as clean.
+- **Display** `(partial — no stored hash)` marker for claims with `span_partial=True` in `format_claims_resolution` (#66).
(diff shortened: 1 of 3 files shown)


Paul-Kyle (approved):
Reviewed against #66. The renderer change is correctly scoped, the positive and negative marker cases are covered, and the Unreleased changelog entry is present. Verified on the exact head commit: 44/44 focused tests, Ruff lint clean, full local suite 2,654 passed (11 skipped, 5 xfailed), and all six hosted checks green.

PR #80, opened 2026-08-03
fix: enforce canonical enums across surfaces

### LINKED ISSUE #69: Enum drift across surfaces — align to the canonical lists in core/parity.py
`palinode/core/parity.py` declares `CATEGORIES`, `MEMORY_TYPES`, and `PROMPT_TASKS` with docstrings describing them as the single source of truth for the four surfaces (MCP, REST, CLI, plugin). Nothing imports them — each surface hand-writes its own literals, and they have drifted: the TypeScript plugin uses singular category names and is missing `ResearchRef` from its type list, and the API's task enum is missing `nightly-consolidation`.

Task: make the Python surfaces import the parity lists instead of repeating them; regenerate the plugin's literals to match (they can't import Python, but a comment pointing at `parity.py` + a drift test closes the loop); extend the existing parity tests to assert each surface's enum equals the canonical list, so the next drift fails CI instead of shipping.

Good first issue: the canonical lists already exist, the parity test file shows the pattern, and every change is mechanical alignment.

WilliamK112:
Closes #69.

## Why

`CATEGORIES`, `MEMORY_TYPES`, and `PROMPT_TASKS` were documented as the canonical cross-surface contract, but each Python surface repeated literals and the TypeScript plugin had already drifted. That left plural categories, `ResearchRef`, and `nightly-consolidation` inconsistent depending on which surface a caller used.

## What changed

- import the canonical tuples in CLI, MCP, and REST schemas
- make REST and CLI category/type choices closed enums, including search type filters
- mirror the canonical category and memory-type literals in the OpenClaw plugin, with an explicit pointer back to `core/parity.py`
- extend the Python parity suite to compare real Click, MCP, and OpenAPI schema values and read the plugin constants using the repository’s established cross-language guard pattern
- update the parity docs, changelog, and one test fixture that used the now-invalid `Note` type

## Behavior note

I checked the API paths for singular categories. They were not accepted as aliases or coerced to plural values: list compared the raw value to the directory name, search forwarded the raw value to storage, and the plugin forwarded its free-form string unchanged. A singular category could therefore be accepted syntactically but produce no matches. This PR makes that failure explicit by rejecting non-canonical category values. Likewise, unknown save types previously fell back to `inbox`; the closed canonical `MEMORY_TYPES` contract now rejects them instead.

## Verification

- `pytest tests/test_surface_parity.py -q` — 208 passed, 1 expected xfail
- focused real-server/stdio tests — 5 passed
- `npm test` — 61 passed
- strict TypeScript check for `plugin/index.ts` — passed
- `ruff check palinode/ tests/ scripts/` — passed
- `bandit -r palinode/ -ll` — passed

No new dependencies.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -63,6 +63,12 @@ reader before writing new records with v0.10.0.
 
 ### Fixed
 
+- **Surface enums now stay aligned with the canonical parity registry** (#69).
+  CLI, MCP, and REST schemas import the category, memory-type, and prompt-task
+  lists from `palinode/core/parity.py`; the OpenClaw plugin mirrors those values
+  with a cross-language drift test. This restores plural category names,
+  includes `ResearchRef`, and exposes `nightly-consolidation` through the API.
+
 - **`quote_hash` is now algorithm-prefixed, so citation anchors can outlive their hash
   function.** Anchors were written as a bare MD5 hex digest, inherited from the dedup
   hasher. A bare digest records no algorithm, so there was no way to tell MD5 from
--- a/docs/PARITY.md
+++ b/docs/PARITY.md
@@ -79,7 +79,9 @@ Singular variants (`person`, `project`, etc.) are **entity-ref prefixes**, not c
 PersonMemory, Decision, ProjectSnapshot, Insight, ResearchRef, ActionItem
 ```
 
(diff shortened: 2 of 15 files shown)


Paul-Kyle (approved):
Reviewed against #69 and the canonical parity contract. The change is correctly scoped across CLI, MCP, REST, and plugin schemas; the new guards inspect the real Click, MCP, OpenAPI, and TypeScript values. Verified on exact head b94e923: 208 parity tests passed with one expected xfail, 61 plugin tests passed, strict production TypeScript and Ruff passed, the full local Python suite passed (2,669 passed, 11 skipped, 5 xfailed), and all six hosted checks are green.

PR #81, opened 2026-08-03
ci: test the unit suite on macOS

### LINKED ISSUE #70: ci: add macOS to the test matrix — the classifier says OS-independent, CI only tests Linux
`pyproject.toml` advertises `Operating System :: OS Independent`, and a large share of the MCP-client audience runs macOS — but CI tests only `ubuntu-latest`. Real macOS differences we currently can't see: `O_NOFOLLOW` semantics (used by the TOCTOU-read security tests), path resolution, filesystem encoding, BSD-vs-GNU tool behavior.

Task: add `macos-latest` to the test matrix in `.github/workflows/ci.yml` for at least one Python version, and fix whatever it flushes out (or file follow-ups for anything non-trivial). Good first issue if the suite passes; a genuinely useful contribution either way.

WilliamK112:
Closes #70.

## Why

Palinode advertises OS-independent support and relies on filesystem behavior that differs on macOS, but the unit suite only ran on Ubuntu. That left macOS path resolution, encoding, and security-sensitive file handling outside the pre-merge signal.

## What changed

- retain the existing Ubuntu Python 3.11 and 3.12 unit-test lanes
- add a macOS-latest unit-test lane on Python 3.12
- drive the unit-test job from an explicit OS/Python include matrix
- update the workflow header so it no longer claims every job runs on Ubuntu

## Verification

- parsed the workflow YAML and verified all three matrix entries
- ran the same non-integration unit-test scope locally on Darwin: 2539 passed, 9 skipped, 5 xfailed
- git diff --check passed

CI-only change: skip-changelog.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,8 @@
 #
 # This replaces the previous ci.yml / main-ci.yml / test.yml trio, whose
 # private-runner jobs could never run here and left every run queued until it
-# expired. All jobs below run on GitHub-hosted ubuntu-latest.
+# expired. All jobs below run on GitHub-hosted runners; unit tests cover Linux
+# plus one macOS lane for filesystem and platform compatibility.
 #
 # All embeddings / LLM calls are mocked in the unit suite — no external
 # services are needed.
@@ -18,11 +19,17 @@ on:
 
 jobs:
   unit-tests:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
-        python-version: ["3.11", "3.12"]
+        include:
+          - os: ubuntu-latest
+            python-version: "3.11"
+          - os: ubuntu-latest
+            python-version: "3.12"
+          - os: macos-latest
+            python-version: "3.12"
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,11 +34,23 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python ${{ matrix.python-version }} (Linux)
+        if: runner.os == 'Linux'
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
+
+      - name: Set up Python 3.12 (macOS — extension-loading build)
+        if: runner.os == 'macOS'
+        run: |
+          brew install python@3.12
+          echo "$(brew --prefix python@3.12)/libexec/bin" >> "$GITHUB_PATH"
+
+      - name: Verify loadable SQLite extensions (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          python -c 'import sqlite3; assert hasattr(sqlite3.connect(":memory:"), "enable_load_extension")'
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,8 @@ jobs:
         if: runner.os == 'macOS'
         run: |
           brew install python@3.12
-          echo "$(brew --prefix python@3.12)/libexec/bin" >> "$GITHUB_PATH"
+          "$(brew --prefix python@3.12)/bin/python3.12" -m venv "$RUNNER_TEMP/palinode-venv"
+          echo "$RUNNER_TEMP/palinode-venv/bin" >> "$GITHUB_PATH"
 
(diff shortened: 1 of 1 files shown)


Paul-Kyle:
Three-lane matrix is exactly right, and thanks for fixing the header claim while you were in there — LGTM. One curiosity for the record: which 5 tests skip on Darwin but not Linux? (2539+9 vs 2544+4 — the totals reconcile, I'd just like them named.) Approving the workflow run now; will merge when the matrix is green.

Paul-Kyle:
The lane earned its keep on run one — but the 315 errors aren't a palinode or PR bug: actions/setup-python's macOS builds compile CPython without loadable SQLite extensions, so sqlite3.Connection.enable_load_extension doesn't exist and the vector store can't load sqlite-vec. Your local run passed because Homebrew/pyenv builds enable it. Suggested fix in this PR — use Homebrew's Python on the macOS lane (the runner image already has brew):

- name: Set up Python (macOS — extension-loading build)
 if: runner.os == 'macOS'
 run: |
brew install python@3.12
echo "$(brew --prefix python@3.12)/libexec/bin" >> "$GITHUB_PATH"

keeping actions/setup-python for the Ubuntu entries (if: runner.os == 'Linux'). Whatever's still red after that — I see a fingerprint-subprocess failure and a uvicorn startup timeout in the log — is the genuine Darwin list, and per the issue those can be follow-up filings rather than blockers here.

Paul-Kyle:
The first macOS lane ran and found a real platform issue rather than a flaky test: the `actions/setup-python` 3.12 build on `macos-latest` exposes an `sqlite3.Connection` without `enable_load_extension`. Palinode reaches that in `get_db()` before `sqlite_vec.load(db)`, which cascades into the 315 setup errors and 34 failures in this run:

https://github.com/phasespace-labs/palinode/actions/runs/30799469047/job/91754220648

That matches the documented sqlite-vec macOS limitation. Please keep the macOS lane, but provision it with a Python/SQLite build that supports loadable extensions (the sqlite-vec docs recommend Homebrew Python) and add a small preflight such as `python -c "import sqlite3; assert hasattr(sqlite3.connect(\":memory:\"), \"enable_load_extension\")"` before the suite. Please do not paper over this by skipping the DB-backed tests—the failure is exactly the portability signal #70 was meant to surface.

Paul-Kyle:
Homebrew was the right move and the verify step is exactly the right guard — the run just died one layer later: Homebrew's Python is PEP 668 externally-managed, so pip refused the install before the suite ever ran. (The Ubuntu reds are only fail-fast cancellations, nothing of yours.)

Fix: build a venv from the brew interpreter and put it on GITHUB_PATH:
```

- name: Set up Python 3.12 (macOS — extension-loading build)
  if: runner.os == 'macOS'
  run: |
    brew install python@3.12
    "$(brew --prefix python@3.12)/libexec/bin/python" -m venv "$RUNNER_TEMP/venv"
    echo "$RUNNER_TEMP/venv/bin" >> "$GITHUB_PATH"
```

Same interpreter, so extension loading survives — your verify step will confirm it. While you're in the matrix: fail-fast: false under strategy:, so a macOS failure can't cancel the Ubuntu baseline again.

Whatever red remains once install succeeds is the genuine Darwin list — exactly what #70 exists to surface. We'll take those as scoped follow-ups, not blockers here.

WilliamK112:
Thanks for the diagnosis and the Homebrew Python pointer. Fixed at `7607a9e`: the macOS lane now installs Homebrew Python 3.12, creates a runner-temp virtual environment (to respect PEP 668), puts that venv on `GITHUB_PATH`, and verifies `enable_load_extension` before installing dependencies.

[The exact-head run](https://github.com/phasespace-labs/palinode/actions/runs/30892914585) is green across all lanes. The macOS lane completed with **2574 passed, 9 skipped, 5 xfailed**; Ubuntu 3.12 completed with **2577 passed, 6 skipped, 5 xfailed**.

For the skip-list record, the current same-head comparison has three macOS-only skips:

- `tests/test_deploy_systemd.py::test_install_sh_shellcheck` — `shellcheck` is absent on the macOS runner.
- `tests/test_deploy_systemd.py::test_install_sh_rejects_system_without_root` — explicitly Linux-only because non-Linux exits at the systemd OS gate first.
- `tests/test_setup_worktree.py::test_shellcheck_clean` — `shellcheck` is absent on the macOS runner.

The earlier “five” inference compared my older local result with a Linux result from a different revision; the current exact-head matrix is the reliable apples-to-apples count, so the correct present-day delta is three. Sorry for the earlier ambiguity.

Paul-Kyle:
Merged — green on the first run after the venv fix, and the exact-head skip accounting (three environmental skips, zero genuine Darwin failures) closes #70 cleaner than it dared ask for. fail-fast: false is folded in maintainer-side and ships with the next sync.

PR #88, opened 2026-08-04
Gate palinode_save inputSchema size directly

### LINKED ISSUE #72: `palinode_save`'s MCP tool schema exceeds some clients' size caps and gets silently dropped
`palinode_save`'s MCP schema is ~4.8 KB. At least one MCP client caps tool schemas at 4,096 bytes and, on exceeding the cap, replaces the schema with an **empty object** rather than truncating or erroring — the tool still appears in the model's tool list, but with no parameter contract, so the model is invited to call a write tool blind.

Task: get the schema under 4 KB without losing the contract. The bulk is description text; candidates include tightening the per-field descriptions, moving worked examples out of the schema into the tool description, and auditing optional fields whose descriptions restate the obvious. Constraint: the descriptions are load-bearing for model behavior (several encode rules like "omit to leave unmarked"), so this is editing, not deleting — proposals in the PR welcome before the full change.

nolanefe:
## Summary

Fixes #72.

Adds a regression gate for the compact serialized `palinode_save` `inputSchema`, matching the schema size that affected MCP clients actually enforce.

Current `main` already contains the description reductions needed to bring the schema to approximately the requested 3.5 KB target, so this PR deliberately avoids further compressing model-facing guidance.

## Size

- Compact `inputSchema`: **3,498 bytes**
- Regression budget: **3,584 bytes**
- Headroom below 4,096-byte client cap: **598 bytes**
- Full tool definition, supplementary: **3,815 bytes**

The regression measurement uses compact JSON serialization of `inputSchema` itself rather than the complete MCP tool definition.

## Changes

- Add a dedicated compact-`inputSchema` size regression test for `palinode_save`
- Keep the existing full-wire-size checks for the other MCP tools
- Document the regression protection in the changelog

No MCP field names, types, requiredness, defaults, validation behavior, or tool semantics change.

## Validation

- `pytest tests/test_mcp_schema_size_budget.py tests/test_mcp_schema_size.py tests/test_mcp_parity.py tests/test_mcp.py -v --tb=short` — **69 passed**
- `pytest tests/ -v --tb=short --ignore=tests/integration` — **2575 passed, 9 skipped, 5 xfailed**
- `ruff check palinode/ tests/ scripts/` — **passed**
- `bandit -r palinode/ -ll` — **passed**
- `git diff --check` — **passed**

Integration tests were not run because they require live integration setup; this change affects only schema-size regression coverage.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to Palinode. Format follows [Keep a Changelog](https://keepa
 
 ### Fixed
 
+- **`palinode_save` MCP schema size regression now gates compact `inputSchema` (#72).** v0.9.7 trimmed description prose and added a per-tool wire-size budget (#765), but that gate measured name + description + inputSchema together. The new `test_palinode_save_input_schema_fits_budget` enforces a ~3.5 KB (3,584 B) limit on the compact serialized `inputSchema` itself — the form clients cap — so schema growth is caught before a client silently replaces the contract with `{}`.
+
 ### Removed
 
 ### Security
--- a/tests/test_mcp_schema_size_budget.py
+++ b/tests/test_mcp_schema_size_budget.py
@@ -17,12 +17,11 @@ an aggregator's own tool accounting can observe it. Staying under a plausible
 client cap is therefore the only lever available here, and this test is the
 only thing that can notice when a new field spends the remaining room.
 
-**When this test fails, split the tool — do not compress prose.** Shrinking
-descriptions buys a few hundred bytes and leaves the next field to re-break it;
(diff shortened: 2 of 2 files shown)


Paul-Kyle:
Merging — this is exactly the gate #72 needed: compact inputSchema (the form clients cap), a 3,584 B budget over the current 3,498 B, with the full-wire belt retained; verified against our tree as well (69 passed, same byte count). One note for the record: the split-don't-compress paragraph your failure message points readers to shouldn't have left the docstring — I'll restore it maintainer-side. The honest scoping — recognizing main had already done the trimming and shipping only the missing gate — is the part I'd hold up as a model.

PR #92, opened 2026-08-09
test: isolate doctor filesystem discovery

### LINKED ISSUE #86: Unit tests that call doctor scan the developer's entire home directory
Doctor's default search roots include the user's home directory. Any test that calls
`doctor.run_all()` without pinning `doctor.search_roots` therefore walks all of `$HOME`:
it can roughly double a local full-suite run (one measurement: 326 s vs 199 s), makes
runtime machine-dependent, and is a plausible contributor to wall-clock-sensitive flakes.

Task: add a conftest default that pins `doctor.search_roots` to `tmp_path` for tests, with
an explicit opt-out for any test that genuinely exercises root discovery. Acceptance: doctor
tests touch no path outside their tmp dir (one meta-test can prove it by counting filesystem
visits under a monkeypatched walker), and full-suite wall-clock on a machine with a large
home directory drops accordingly.

WilliamK112:
Closes #86

## Why

An empty `config.doctor.search_roots` activates the production fallback, which includes the user's entire home directory. Tests that invoke doctor through the process-wide config were therefore machine-dependent, could inspect real developer databases, and paid an unbounded filesystem-walk cost unrelated to what they were testing.

## Changes

- Pin the process-wide doctor search roots to each test's `tmp_path` with an autouse fixture.
- Register a narrow `doctor_real_search_roots` marker for tests that intentionally exercise production root discovery.
- Explicitly isolate the diagnostics-skeleton helper, which constructs its own `Config` and therefore bypasses the process singleton.
- Add regression coverage that records every `os.walk` root and proves the default doctor check visits exactly the current test temp directory, plus coverage for the marker opt-out.

## Validation

- `/Users/William/.local/bin/uv run --extra dev pytest -q` — 2798 passed, 10 skipped, 5 xfailed
- `/Users/William/.local/bin/uv run --extra dev ruff check palinode/ tests/ scripts/` — passed
- `/Users/William/.local/bin/uv run --extra dev bandit -r palinode/ -ll` — no medium/high findings
- `git diff --check` — passed

## Changelog

`skip-changelog`: this changes test isolation only and does not alter shipping behavior.

## AI disclosure

This change was implemented and validated by an AI coding agent working under the WilliamK112 account.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,6 +115,7 @@ addopts = "--ignore=tests/live"
 pythonpath = ["."]
 markers = [
     "slow: marks tests that need Ollama or take >5s",
+    "doctor_real_search_roots: allow a test to exercise doctor's built-in filesystem root discovery",
 ]
 asyncio_mode = "strict"
 timeout = 60
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -81,6 +81,33 @@ def _isolate_global_config():
 
 
 @pytest.fixture(autouse=True)
+def _isolate_doctor_search_roots(
+    request: pytest.FixtureRequest,
+    tmp_path: pathlib.Path,
+    monkeypatch: pytest.MonkeyPatch,
+    _isolate_global_config,
+) -> None:
+    """Keep doctor filesystem discovery inside each test's temp directory.
+
+    An empty ``config.doctor.search_roots`` enables the production fallback,
+    which includes the user's entire home directory. Tests that invoke doctor
+    through the process-wide config must never inherit that machine-dependent
+    behavior: it is slow, can discover a developer's real databases, and makes
+    wall-clock results depend on the size of ``$HOME``.
+
+    A test that specifically covers the production fallback may opt out with
+    ``@pytest.mark.doctor_real_search_roots``. Keep that marker narrow: ordinary
+    doctor behavior tests should provide explicit roots instead.
+    """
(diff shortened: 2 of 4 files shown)


Paul-Kyle:
Green across all seven lanes, including the macOS one you added — merging.

The opt-out is the right shape: `get_closest_marker` on the node with an early return, so
the marker costs nothing at runtime and needs no teardown branch.

One thing left behind. `test_diagnostics_skeleton.py`'s `_ctx` was one of two synthetic-
`Config` helpers; the other is `_ctx` in `tests/test_doctor_path_checks.py`, which takes
`search_roots: list[str] | None = None` and sets it only when non-None. It isn't leaking
today — every `phantom_db_files` call site in that file passes roots explicitly — but it's
the same latent bug you just closed, and the autouse fixture can't reach it. Flipping the
default closes the class rather than the instance:

```python
    cfg.doctor.search_roots = (
        search_roots if search_roots is not None else [str(memory_dir)]
    )
```

...leaving root-discovery tests to pass their roots explicitly. Follow-up or a second
commit here, your call — not a blocker.

Non-blocking, and only so you know the failure mode if it ever bites: patching
`phantom_db.os.walk` patches the real `os` module, so `visited_roots == [tmp_path.resolve()]`
also fails if anything else in-process walks during that window — and this suite does keep
background timers alive. If that test ever flakes, that's the reason, not your assertion.

Issue #93, opened 2026-08-10
Native Windows: test suite fails on directory fsync and locale-default file encoding


Paul-Kyle:
Guan Tong (@happykawayigt) surfaced native-Windows test failures while working on #82. I
recall five failures, but the original failure list wasn't captured, so filing this on his
behalf keeps the finding and credit on the tracker. The specific mechanisms below come from
a follow-up code read, not from his report; treat the mapping between them and the original
failures as unconfirmed until someone reproduces on Windows.

Native Windows is not in CI today (Docker is the supported Windows path), so nothing here
affects shipped behavior. It does mean the suite cannot currently be run natively on
Windows, which closes the door on Windows contributors.

## Two mechanisms found by inspection

**1. `_fsync_directory` opens a directory as a file descriptor.**
`palinode/core/git_tools.py:87`

```python
def _fsync_directory(path: str) -> None:
    dir_fd = os.open(path, os.O_RDONLY)
```

Windows does not permit opening a directory this way through `os.open`; the expected
failure is `PermissionError: [WinError 5] Access is denied`. This sits in
`write_memory_file`, the single write primitive for every memory-file mutation, so a native
Windows test that saves through the real path reaches it.

The fix is a portability decision, not merely a test skip: the directory fsync exists so a
rename survives a crash, and Python exposes no direct directory-fsync equivalent on
Windows. Prefer an explicit `os.name != "nt"` guard with a comment documenting the weaker
durability guarantee on Windows; do not hide a silent no-op inside a function whose name
claims it flushed the directory.

**2. Test helpers call `open()` without an encoding.**
`tests/test_source_capture_g1.py:40`

```python
with open(file_path, "r") as f:
```

With no `encoding=`, Python uses the locale default — often a legacy Windows code page such
as cp1252 unless UTF-8 mode is enabled — to read files Palinode wrote as UTF-8. That file's
fixtures include non-ASCII content by design (line 76 contains smart quotes), so the
round-trip can compare mojibake against the original. `_frontmatter` is a shared helper,
which is consistent with a small cluster of failures in one file.

The pattern is broader than that helper. A simple same-line grep on current public `main`
finds **149** bare `open(` lines without `encoding=` across **36** top-level test files, or
**167 across 41 files** when `tests/integration/` is included. This is a heuristic inventory,
not an AST result — multiline calls can be misclassified — and most sites are harmless
today because their fixtures are ASCII. The acceptance scope is narrower: helpers that read
[...]

PR #94, opened 2026-08-10
Native Windows: test suite fails on directory fsync and locale-default file encoding

deepunyk:
Addresses #93

Implemented the focused Windows portability fix for [issue #93](https://github.com/phasespace-labs/palinode/issues/93):

- Skips unsupported directory fsync on Windows.
- Falls back from unavailable `os.fchmod` on Python 3.11/3.12.
- Forces UTF-8 in source-capture tests and adds non-ASCII coverage.
- Added changelog entry.

Checks passed: `git diff --check`, AST parsing, compileall, and a Windows portability smoke test.

`pytest`, `ruff`, and `bandit` were unavailable; setup failed because `ensurepip` is missing. No typecheck command is configured. Native Windows CI remains unverified. No commit created.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ All notable changes to Palinode. Format follows [Keep a Changelog](https://keepa
 
 ### Fixed
 
+- **Native Windows memory writes now avoid unsupported directory fsync and
+  preserve overwrite permissions on Python 3.11/3.12.** Source-capture tests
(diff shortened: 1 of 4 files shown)


Paul-Kyle (approved):
Validated the exact head locally: 31 focused tests passed, then the full public suite passed (2,799 passed, 11 skipped, 5 expected xfails). All seven public checks are also green. Merging this focused fix while keeping #93 open until the suite is actually exercised on native Windows.

PR #95, opened 2026-08-10
ci: upload security scan reports

### LINKED ISSUE #71: ci: upload bandit + pip-audit reports as artifacts on failure
The security-scan job runs bandit and pip-audit but surfaces only exit codes — a contributor whose PR fails the scan has to reproduce locally to see why. Task: emit both reports as JSON (`bandit -r palinode/ -ll -f json -o …`) and upload with `actions/upload-artifact` under `if: always()`, so the failure page carries the findings. CI-yaml only; no product code.

WilliamK112:
Closes #71

## Why
The security scan preserved scanner exit status but not findings. Contributors had to reproduce CI locally to inspect Bandit and pip-audit results.

## Changes
- write Bandit results to `bandit-report.json`
- write pip-audit results to `pip-audit-report.json`
- upload both as `security-scan-reports` under `if: always()`
- keep the existing `continue-on-error` behavior and warn if a scanner cannot produce a file

## Validation
- workflow YAML parsed successfully
- actual Bandit JSON command produced valid JSON (47,588 bytes)
- actual pip-audit JSON command produced valid JSON (67,419 bytes), including when it exited 1 for findings
- full pytest: 2796 passed, 10 skipped, 5 xfailed
- full Ruff check passed
- `git diff --check` passed

## Changelog
`skip-changelog`: CI-only change

## AI disclosure
This change was implemented and validated by an AI coding agent working under the WilliamK112 account.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,11 +91,19 @@ jobs:
           pip install bandit pip-audit
 
       - name: Run bandit (static security analysis)
-        run: bandit -r palinode/ -ll -f json -o bandit-report.json
+        run: |
+          status=0
+          bandit -r palinode/ -ll || status=$?
+          bandit -r palinode/ -ll -f json -o bandit-report.json || status=$?
+          exit "$status"
         continue-on-error: true
 
       - name: Run pip-audit (dependency vulnerability check)
-        run: pip-audit -f json -o pip-audit-report.json
+        run: |
+          status=0
+          pip-audit || status=$?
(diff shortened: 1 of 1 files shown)


WilliamK112:
Thanks for catching that. I updated the PR in de41d02 so each scanner still runs in its existing human-readable mode for the job log, then runs its JSON output mode for the artifact. The shell block carries a nonzero status from either invocation through the existing advisory `continue-on-error` step, so the artifact is additive without weakening the scan semantics.

PR #98, opened 2026-08-10
test: make live search result self-contained

### LINKED ISSUE #96: tests/live: `test_search_returns_results` asserts content the test never saved
`tests/live/test_live_instance.py::TestSearch::test_search_returns_results` searches for a
fixed phrase and asserts at least one hit:

```python
resp = client.post("/search", json={"query": "palinode memory", "limit": 3})
assert len(results) > 0, "Search returned no results"
```

Nothing in the test puts anything matching "palinode memory" into the instance. It passes only
when the instance happens to already hold related content, so the same test passes or fails
depending on which machine it runs against — and against a freshly-started instance holding
only the unrelated documents the preceding test wrote, a relevance threshold correctly filters
everything out and it fails.

It used to pass for the wrong reason: the threshold previously filtered a fused score that was
a function of rank rather than relevance, so in a small store nothing was ever filtered.
Now that the threshold filters on per-arm relevance, the test says what it always meant to say
— and what it means is wrong.

Task: make the test self-contained. Save content the test itself generates, with a unique
identifiable token in it, then query for *that* token and assert it comes back. The test
should then pass against an empty instance and a full one alike, and it should fail for a real
reason — search is broken — rather than because of what the instance happened to contain.

Do not solve this by skipping when the store is empty. The failing case here is a store that
is *non-empty but irrelevant*, which a skip doesn't cover, and a test that skips itself on the
most common local setup is a test nobody runs.

Acceptance: the test creates its own searchable content and asserts on it; it passes from a
cold start; and no assertion depends on pre-existing instance content.

floze-the-genius:
## Why

The live search smoke searched for fixed ambient content that it never created, so its outcome depended on what happened to be stored in the target instance.

Closes #96.

## What changed

- save a memory containing a per-run marker before searching
- use the `/save` inline-index result to search immediately when indexing completed
- fall back to a bounded 30-second watcher poll only when inline indexing was deferred, with the first search before any sleep
- include `indexed`, `embedded`, and `index_error` from the save response in failures
- document the test correction under Unreleased

## Validation

- fresh local API with 0 files / 0 chunks and an Ollama-compatible HTTP embedding endpoint: focused live test passes immediately (1 passed in 0.19s)
- isolated control-flow validation: inline/no-sleep, deferred retry, and 30-second diagnostic failure paths pass
- full default suite: 2804 passed, 8 skipped, 5 xfailed
- `bandit -r palinode/ -ll` (0 medium/high findings)
- `git diff --check`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,10 @@ All notable changes to Palinode. Format follows [Keep a Changelog](https://keepa
 
 ### Fixed
 
+- **The live search smoke test is now self-contained.** It saves and queries a
+  unique marker instead of depending on unrelated content already being present
+  in the tested instance.
+
 - **Native Windows memory writes now avoid unsupported directory fsync and
   preserve overwrite permissions on Python 3.11/3.12.** Source-capture tests
   read Palinode-written files as UTF-8 instead of using the locale default.
--- a/tests/live/test_live_instance.py
+++ b/tests/live/test_live_instance.py
@@ -159,11 +159,28 @@ class TestSearch:
 
     @pytest.mark.slow
     def test_search_returns_results(self):
-        """Search for something that should exist in any palinode instance."""
-        resp = client.post("/search", json={"query": "palinode memory", "limit": 3})
-        assert resp.status_code == 200
-        results = resp.json()
-        assert len(results) > 0, "Search returned no results"
+        """Search returns content created by this test, not ambient instance data."""
+        unique = f"xyzzy-{_TEST_PREFIX}-search-result"
+        save_resp = client.post("/save", json={
+            "content": f"Live test: {unique}. This phrase belongs to the search result test.",
+            "type": "Insight",
+            "slug": f"{_TEST_PREFIX}-search-result",
+        })
+        assert save_resp.status_code == 200
+
+        # Poll for up to 30s — watcher indexing is asynchronous.
+        for _attempt in range(6):
+            time.sleep(5)
+            resp = client.post("/search", json={"query": unique, "limit": 3})
+            assert resp.status_code == 200
+            results = resp.json()
+            if any(unique in result.get("content", "") for result in results):
+                break
+        else:
+            pytest.fail(
+                f"Search did not return the memory containing '{unique}' after 30s — "
+                "watcher may not be running or Ollama unreachable"
+            )
 
     @pytest.mark.slow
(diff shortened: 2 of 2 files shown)


Paul-Kyle:
Approved the workflow run. The substance here is right — saving a per-run marker and querying
for that exact marker is exactly what #96 was after, and the changelog bullet correctly lands
under the existing `### Fixed` heading rather than adding one.

One thing to change before this goes in. The loop sleeps before its first query:

```python
for _attempt in range(6):
    time.sleep(5)
    resp = client.post("/search", ...)
```

so every run costs 5 seconds even when the content was searchable immediately — which is why
your focused run took 5.48s, essentially all of it sleeping.

The comment says watcher indexing is asynchronous, and that's the wrong model for this path.
`/save` indexes **inline**: it calls the indexer synchronously inside the request and returns
`indexed`, `embedded`, and — if inline indexing failed — `index_error`. The watcher is the
fallback for filesystem-direct writes and for retrying a failed inline index, not the primary
path for an API save.

So rather than moving the sleep to the end of the loop, use what the save response already
tells you: if `indexed` (or `embedded`) is true, query immediately and skip polling entirely;
only fall back to the retry loop when inline indexing deferred, with the sleep *after* each
attempt so the first query is never delayed. That makes the test deterministic rather than
merely patient.

While you're there: the failure message guesses ("watcher may not be running or Ollama
unreachable") at something the save response would have told you outright. Include the save
response's `indexed` / `index_error` in the failure text so the test diagnoses itself.

Push those and I'll re-approve the run — merging on green.

floze-the-genius:
Addressed in `675ade0`.

The test now reads the `/save` response, searches immediately with no sleep when `indexed` or `embedded` is true, and uses the watcher fallback only for deferred indexing. The fallback searches at 0/5/10/15/20/25/30 seconds, and failures include `indexed`, `embedded`, and `index_error`.

Fresh-instance validation now completes the focused live test in 0.19s. The inline/no-sleep, deferred-retry, and diagnostic-failure branches pass independently; the full suite is 2,804 passed / 8 skipped / 5 expected failures, with Ruff and Bandit green. The replacement CI and Wheel Smoke runs are waiting for fork approval again:

- https://github.com/phasespace-labs/palinode/actions/runs/31406846878
- https://github.com/phasespace-labs/palinode/actions/runs/31406846894

potatoiscompiled (approved):

PR #102, opened 2026-08-10
fix: use parser.split_frontmatter in cli/read.py (#100)

### LINKED ISSUE #100: `palinode read` mangles the body when frontmatter contains `---`
`palinode/cli/read.py:94-101`:

```python
def _strip_frontmatter(content: str) -> str:
    """If `content` starts with `---`, drop the frontmatter block."""
    if not content.startswith("---"):
        return content
    parts = content.split("---", 2)
    if len(parts) < 3:
        return content
    return parts[2].lstrip("\n")
```

This splits on the first two occurrences of `---` **anywhere in the string**, not on fence
lines. When a frontmatter *value* contains that sequence — a title, a quote anchor, an
em-dash-heavy `description` — the split lands in the wrong place and `palinode read` returns a
truncated, mangled body.

Worth being precise about the trigger, because the obvious guess is wrong: a `---` thematic
break in the **body** is harmless, since `maxsplit=2` leaves the remainder intact. The failure
needs `---` inside the **frontmatter block**.

This is also the third frontmatter-splitting implementation in the tree, and the only one that
is neither the `frontmatter` library nor the canonical helper. `palinode/core/parser.py`
already has `split_frontmatter`, documented as lossless — `frontmatter_block + body == content`
always.

Task: have `cli/read.py` call `parser.split_frontmatter` instead of hand-splitting.

Acceptance: a memory whose frontmatter value contains `---` reads back through `palinode read`
with an intact body, pinned by a test; the hand-rolled `_strip_frontmatter` is gone rather than
patched.

Scope: only `cli/read.py`. The other call site that hand-rolls this is a separate cleanup and
is not part of this issue.

HeaTTap:
Fixes #100

Replaces hand-rolled _strip_frontmatter in cli/read.py with parser.split_frontmatter.

Why:
The previous implementation in cli/read.py split on the first two occurrences of three dashes anywhere in the string. When frontmatter contained three dashes inside a value, the split landed inside the frontmatter block and mangled the body.

Using parser.split_frontmatter uses the proper regex boundary matching frontmatter delimiters, leaving the body intact.

Changes:
- Replaced _strip_frontmatter in cli/read.py with parser.split_frontmatter.
- Added tests in tests/test_cli_read_frontmatter.py covering format_with_meta and CLI read.
- Added CHANGELOG entry under Unreleased Fixed section.

Verification:
- Ran pytest (2801 passed).
- Ran ruff check and bandit (all checks passed).

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,9 @@ All notable changes to Palinode. Format follows [Keep a Changelog](https://keepa
 
 ### Fixed
 
+- **`palinode read` frontmatter stripping.** Using `parser.split_frontmatter`
+  instead of splitting on the first two `---` occurrences preserves the body when
+  frontmatter values contain `---` sequences.
 - **Native Windows memory writes now avoid unsupported directory fsync and
   preserve overwrite permissions on Python 3.11/3.12.** Source-capture tests
   read Palinode-written files as UTF-8 instead of using the locale default.
--- a/palinode/cli/read.py
+++ b/palinode/cli/read.py
@@ -16,6 +16,7 @@ import click
 
 from palinode.cli._api import HTTPStatusError, api_client
 from palinode.cli._format import OutputFormat, get_default_format
+from palinode.core.parser import split_frontmatter
 
 
 @click.command()
@@ -86,16 +87,6 @@ def _format_with_meta(result: dict) -> str:
     # When meta=True, the API returns the full file content (frontmatter +
     # body).  Strip the leading frontmatter block for cleaner CLI output.
     content = result.get("content", "")
-    body = _strip_frontmatter(content)
+    _, body = split_frontmatter(content)
     lines.append(body)
     return "\n".join(lines)
-
-
-def _strip_frontmatter(content: str) -> str:
-    """If `content` starts with `---`, drop the frontmatter block."""
-    if not content.startswith("---"):
-        return content
-    parts = content.split("---", 2)
-    if len(parts) < 3:
-        return content
-    return parts[2].lstrip("\n")
--- a/tests/test_cli_read_frontmatter.py
+++ b/tests/test_cli_read_frontmatter.py
@@ -1,0 +1,69 @@
+"""Unit tests for `palinode read` frontmatter handling in `cli/read.py`.
+
+Issue #100: `palinode read` was mangling the body when frontmatter values contained `---`.
+Replacing the hand-rolled `_strip_frontmatter` helper with `parser.split_frontmatter`
+ensures the frontmatter block is cleanly stripped based on YAML boundary regex rather than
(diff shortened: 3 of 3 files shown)


Paul-Kyle:
Thanks — the implementation is right. This CI failure is ours: the guard cannot distinguish tracker context in a bare `#100`, and its “private issue” diagnosis is wrong for a public issue you are correctly fixing. Please change only the module docstring to use the qualified URL, `Issue https://github.com/phasespace-labs/palinode/issues/100:`, or name the defect without a number. No implementation change is needed. Push that one-line revision and I will reapprove the workflows and merge on green.

Paul-Kyle:
An update and an apology. The same fix landed on our side a few hours before I opened #100, and
I didn't check before asking you to take it on. That's our process failing, not anything you
did.

Your change is still the right one for this repository — the internal fix hasn't reached here
yet — so I'd like to merge this rather than close it. If you're still up for pushing the
docstring adjustment, I'll merge on green and the credit is yours. If you'd rather not spend more
time on it now, that's completely fair: say so and I'll finish it on your commit with your
authorship preserved.

Paul-Kyle:
No pressure at all — just flagging that this is still open and still wanted whenever you have a
moment. The docstring line is the only thing standing between it and a merge.

And if you'd rather hand it back, that's genuinely fine too: say the word and I'll finish it on
your commit with your authorship kept.

Paul-Kyle:
I owe you a straight answer and an apology, and the answer is that this is finished — but not by
either of us.

The fix reached public `main` on 2026-08-11 at 03:58Z, in the v0.12.0 release sync. Our development
happens in a private repo and lands here as a squashed release, and that release carried the
internal version of this same fix. `palinode/cli/read.py` on `main` now imports `split_frontmatter`
and uses it, which is exactly the change you made. I confirmed the bug no longer reproduces: a `---`
inside a frontmatter value leaves the body intact and the thematic break survives.

The part I am sorry about is the timing. I told you at 03:28Z that the internal fix hadn't reached
here yet — true when I wrote it. It landed at 03:58Z. And at 04:24Z I asked you again for the
docstring line, saying it was the only thing between this and a merge. That was already wrong, and I
didn't check before sending it. You then spent four days holding a patch that had stopped mattering
half an hour earlier.

None of that is a reflection on the work. You diagnosed it correctly, you fixed it the right way —
the canonical helper rather than a second hand-rolled splitter — and you were the first person to
open a PR against it. The three things that delayed you were all ours: a CI guard that couldn't read
a bare `#100`, an issue we seeded without checking our own tree, and now a release overlay we didn't
account for.

So I'm closing this. Your diagnosis and implementation were right, and you should have had a clean
review or a prompt explanation — not four days of silence around a patch the release had already
made unnecessary. I'm sorry we handled it this way.

Issue #101, opened 2026-08-10
The `recall:` config section is documented and shipped but read by nothing


Paul-Kyle:
```
$ grep -rn 'config\.recall' palinode/ | wc -l
0
```

Zero readers — and yet `recall:` is **60 lines of the shipped
`palinode.config.yaml.example`** (from line 29) and is documented in the README. Users are
being invited to tune knobs that cannot affect anything.

The TypeScript plugin doesn't rescue it either: `plugin/index.ts` never loads
`palinode.config.yaml`. It hardcodes every equivalent value — the refresh interval, the minimum
query length, the trivial-query patterns are all inline literals. One of the documented
settings appears in the plugin *only inside a comment*.

This is residue from the pre-[ADR-001](ADR-001-tools-over-pipeline.md) injection pipeline. That
ADR calls the four-phase pipeline "compensating complexity — disposable"; the pipeline was
disposed of and its configuration was not.

Task: delete the `recall:` config classes, the 60-line block in the shipped example, and the
README documentation for it.

**Removal is safe for existing users** — worth stating plainly, because deleting shipped config
looks scarier than it is. Config loading ignores unknown keys rather than rejecting them, so a
user whose `palinode.config.yaml` still contains a `recall:` block keeps starting normally
after the upgrade. Their block stops being read, which is already true today. Please verify
that yourself rather than taking it on faith, and say in the PR how you checked.

Acceptance: `grep -rn 'config\.recall' palinode/` still returns nothing (because the readers
never existed), no `recall` classes remain in the config module, the shipped example and README
no longer document it, and the suite passes.

Scope: `recall:` only. Other config sections have their own histories and are not part of this.

Rithikmahadev12:
I can address this by removing the unused `recall:` configuration classes, excising the sixty-line example block, stripping the corresponding README documentation, and verifying that no residual imports or references remain. I am prepared to complete this cleanup for an estimate of ten US dollars, contingent on the scope not being larger than what is visible in the issue alone. Please confirm whether you would like me to proceed, and I will deliver the pull request within a day or two of receiving your approval. I remain available to assist should you wish to move forward.

Paul-Kyle:
Thanks for reading these properly — the plan you've outlined on each one is accurate, which is
more than a skim would produce.

We're not able to take you up on it, though: Palinode doesn't purchase contributions, and there's
no bounty behind these issues. The `good first issue` label here means "a small, well-specified
[...]

Issue #99, opened 2026-08-10
parity: MCP inventory counts the deprecated `palinode_timeline` alias as its own capability


Paul-Kyle:
`palinode_timeline` is an unambiguous deprecated alias of `palinode_history` — declared in
`palinode/mcp.py`, deprecation-logged when called, and its output prefixed `[DEPRECATED]`.
But `palinode/core/parity.py` gives it its own `INVENTORY_BACKLOG` row (line 661) alongside
`palinode_history` (line 654), so the inventory guard counts a single capability twice.

PR #89 established the rule for the CLI surface while adding the `dream` alias: **an alias is
not a capability.** The same object registered under a second name dedups to its canonical
entry rather than reading as new surface area. This is that rule applied to the MCP inventory.

Task: make `palinode_timeline` an alias annotation on `palinode_history`'s entry instead of a
separate row, so the inventory reports one capability with two names rather than two
capabilities.

Acceptance: the inventory count drops by one and reflects real capabilities; the deprecation
behaviour of `palinode_timeline` is unchanged (it still works, still logs, still prefixes
`[DEPRECATED]`); and a test pins that an alias contributes no new inventory row, so the next
alias inherits the rule instead of re-litigating it.

Rithikmahadev12:
I can resolve this by attaching an alias annotation to the existing `palinode_history` inventory entry rather than creating a separate row for `palinode_timeline`, thereby preserving the established deprecation logging and output formatting. I am prepared to implement this adjustment for an estimated $10 USD, contingent upon the scope remaining as contained as the issue description indicates. Please confirm whether this approach meets your requirements before I proceed, and I will submit the patch within a day or two of receiving your approval. I extend this offer as an independent contributor and await your confirmation prior to commencing any work.

Paul-Kyle:
Thanks for reading these properly — the plan you've outlined on each one is accurate, which is
more than a skim would produce.

We're not able to take you up on it, though: Palinode doesn't purchase contributions, and there's
no bounty behind these issues. The `good first issue` label here means "a small, well-specified
way into the codebase", not a listing. That's about how the project is set up rather than
anything about your offer, which was clear and courteous.

If you'd like to work on one of them unpaid you're genuinely welcome. I'd suggest picking a
single issue to start rather than all three — comment on the one you want and it's yours. One
landing well is worth more to both of us than three sitting claimed.

Leaving all three open and unassigned for now.

WilliamK112:
[...]

Issue #83, opened 2026-08-03
lint: flag relative time expressions in memory bodies before they rot


Paul-Kyle:
Relative dates rot. "Yesterday we decided X" is uninterpretable three weeks later and actively
misleading once consolidation merges it into a living document. Consolidation can eventually
rewrite these to absolute dates — a separate, prompt-side change. The first step is making the
drift visible.

## Task

Add a `check_relative_dates(body: str) -> list[dict[str, str]]` function to
`palinode/core/lint.py`, next to `check_wiki_drift` (`:294`), returning one dict per match with
`line` (1-based) and `expression` (the matched text).

Wire it into `run_lint_pass()` the same way `wiki_drift` is wired: accumulate into a
`relative_dates` list inside the per-file loop (see `:529-533`, and follow the `daily/` skip — a
daily note is dated by its filename, so "yesterday" in one is anchored and must not flag), and add
the key to the returned dict (`:634`). Then render it in `palinode/cli/lint.py` alongside the
other checks — a check that is computed but not rendered is invisible to everyone who does not
pass `--format json`.

Report-only, like every lint class. Do **not** try to resolve what the absolute date should be;
that is consolidation's job later.

## Scope — this list, not a general grammar

Match case-insensitively, word-boundary anchored:

    yesterday · today · tomorrow · last week · last month · last year
    next week · next month · next year · this week · this month
    recently · lately · currently · right now · these days
    N days ago · N weeks ago · N months ago · N years ago  (N as digits or one..twelve)

That is the deliverable. A follow-up can widen it; treat anything outside this list as out of
scope for this issue.

## Decisions already made — you do not need to litigate these

- **Quoted text IS flagged.** A relative date inside a quotation rots the same way, the check is
  report-only, and detecting quotation reliably in markdown costs more than the false positives
  are worth. Pin it with a test so it stays deliberate.
- **`daily/` files are skipped**, per the reasoning above.
- **No severity levels.** One flat list of findings.

## Tests

`tests/test_lint.py`, table-driven. Positives covering every entry in the list above. Negatives
that must not flag: an absolute date (`2026-08-03`), a bare month name (`August`), `todays`
inside a longer word, and a `daily/` file containing "yesterday".

Paul-Kyle:
Rewritten. This sat for over a week and that was the issue's fault, not anyone's: it asked whoever
picked it up to decide how quoted text should be handled, which is my call to make, and it named
[...]

Issue #67, opened 2026-08-03
Document the algorithm-prefixed quote_hash format for hand-written frontmatter


Paul-Kyle:
As of 0.10.0, `quote_hash` values are written as `<algorithm>:<hex>` (for example, `sha256:…`), with bare legacy digests resolved by length: 32 hexadecimal characters means MD5 and 64 means SHA-256.

Users who hand-write `sources:` anchors in their memory files have no documentation describing the format.

## Task

Add a short section to the memory-format documentation — `docs/HOW-MEMORY-WORKS.md` or `docs/GIT-MEMORY.md`, whichever fits better — covering:

- the algorithm-prefixed format
- the normalization rule: smart punctuation is folded and whitespace is collapsed
- omitting the hash is valid because it is computed on save

## Why this is a good first issue

This is documentation only. The relevant behavior is documented in `palinode/core/quote_verify.py`.

WilliamK112:
I’d like to take this one. I’ll add a concise section to the most appropriate memory-format guide covering the algorithm-prefixed `quote_hash` format, the smart-punctuation/whitespace normalization rules, and the fact that omitting the hash is valid because Palinode computes it on save. I’ll verify the wording against `palinode/core/quote_verify.py` before opening the PR.

Paul-Kyle:
Yours — and sorry for the order of this. You asked at 09:10 and I'm answering after you'd already
opened #111, which is mine to fix, not yours.

Assigning it so it reads as taken. Reviewing #111 now.

PR #104, opened 2026-08-10
tests/live: make test_search_returns_results self-contained

### LINKED ISSUE #96: tests/live: `test_search_returns_results` asserts content the test never saved

potatoiscompiled:
Fixes #96 by making the test save its own unique searchable string before searching for it.

### DIFF
--- a/tests/live/test_live_instance.py
+++ b/tests/live/test_live_instance.py
@@ -159,11 +159,26 @@ class TestSearch:
 
     @pytest.mark.slow
     def test_search_returns_results(self):
-        """Search for something that should exist in any palinode instance."""
-        resp = client.post("/search", json={"query": "palinode memory", "limit": 3})
-        assert resp.status_code == 200
-        results = resp.json()
-        assert len(results) > 0, "Search returned no results"
+        """Search for a unique phrase that the test itself saves."""
+        unique_phrase = f"unique-search-results-{_TEST_PREFIX}"
+        client.post("/save", json={
+            "content": f"Live test ensuring search returns results: {unique_phrase}",
+            "type": "Insight",
+            "slug": f"{_TEST_PREFIX}-search-returns",
+        })
+        
+        # Poll for up to 30s to allow for indexing
+        found = False
+        for _attempt in range(6):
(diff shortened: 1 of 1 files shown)


Paul-Kyle:
Thank you for this, and I'm sorry you spent the time — #96 was already assigned, and the fix is
in flight as #98, which has been through a review round. So I'm going to close this one as a
duplicate rather than leave it hanging.

That's a gap on our side: the note asking people to claim an issue before starting isn't in
CONTRIBUTING yet. It will be shortly, and it's there because of situations exactly like this.

If you'd like something to pick up, #101, #99, #83 and #67 are all unclaimed right now. Comment
on whichever appeals and it's yours — genuinely, I'd like you to have one.

PR #108, opened 2026-08-11
fix: preserve read bodies with frontmatter hyphens

### LINKED ISSUE #100: `palinode read` mangles the body when frontmatter contains `---`

WilliamK112:
## Why

The read presenter split on any occurrence of three hyphens, so a sequence inside a frontmatter value was mistaken for the closing fence and YAML fragments leaked into the displayed body.

## What changed

- use the canonical fence-line-aware split_frontmatter helper
- remove the hand-rolled CLI splitter
- add Click-level regression coverage for read --meta --format text
- document the fix under Unreleased

Fixes #100.

## Validation

- regression test failed before the implementation and passes after it
- 53 related read/parser tests passed
- full sandbox suite: 2823 passed, 10 skipped, 5 xfailed; the four socket/home-path tests blocked by sandbox policy passed when rerun with those capabilities
- ruff check palinode/ tests/ scripts/
- bandit -r palinode/ -ll
- git diff --check

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,10 @@ All notable changes to Palinode. Format follows [Keep a Changelog](https://keepa
 
 ### Fixed
 
+- **`palinode read --meta --format text` now preserves bodies when a frontmatter
+  value contains `---`.** The CLI uses the canonical fence-line-aware splitter
+  instead of treating any three hyphens inside the YAML block as its closing fence.
+
 - **Concurrent consolidation runs now refuse instead of racing over store
   mutations and git commits.** Weekly and nightly runs share a store-local,
   stale-recoverable lock across API, CLI/MCP, and cron entry points.
--- a/palinode/cli/read.py
+++ b/palinode/cli/read.py
@@ -16,6 +16,7 @@ import click
 
 from palinode.cli._api import HTTPStatusError, api_client
 from palinode.cli._format import OutputFormat, get_default_format
+from palinode.core.parser import split_frontmatter
 
 
 @click.command()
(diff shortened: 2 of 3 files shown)


Paul-Kyle:
Thank you for this — and I'm sorry, because I'm going to hold it rather than merge it, for a
reason that has nothing to do with the work.

HeaTTap opened #102 against the same issue yesterday afternoon. We'd invited him to take it and
then, through a failure entirely on our side, left him waiting a long time on a red build. He's
owed the chance to land it, and merging this would take that away after we'd already cost him an
afternoon.

Nothing about how you approached this was wrong. You asked on the issue first, which is more
than I ask of anyone, and I hadn't answered yet — that's mine, not yours.

I'm leaving this open rather than closing it, so if #102 doesn't come back I'll come to you. And
if you'd like something else in the meantime, #99, #83 and #67 are genuinely open — I've checked
those against the current code rather than assuming.

Paul-Kyle:
Closing this one, and the reason is a good outcome rather than a rejection.

I held it as the fallback in case #102 went quiet. It turns out neither was needed: the fix reached
`main` on 2026-08-11 in the v0.12.0 release sync, because our private tree already had the same
change and the release carried it here. I confirmed the bug no longer reproduces on `main`.

So the hold I put on this was already unnecessary when I wrote it — I hadn't checked what the
release had just changed underneath both PRs. Nothing about your work was the problem, and you asked
on the issue first, which is more than I ask of anyone.

Thanks for the patience on this one. #124 landing today is the better use of the same afternoon.

Issue #113, opened 2026-08-11
palinode lint silently drops two checks from its default output


Paul-Kyle:
`palinode lint` runs two checks whose results never reach the default output.

`run_lint_pass()` computes and returns `wiki_drift` (`palinode/core/lint.py:531-533`, returned at
`:634`) and `missing_priority` (`:482`, returned at `:633`). `palinode/cli/lint.py` renders every
other key in that dict — `missing_fields`, `orphaned_files`, `stale_files`, `contradictions`,
`missing_entities`, `missing_descriptions`, `source_anchor_issues`, `claim_anchor_issues`,
`stale_open_questions`, `open_contradictions`, `entity_aliases`, `core_count` — and neither of
these two.

The result is not a missing warning line, it's a wrong answer. `wiki_drift` enforces the
frontmatter-`entities:`/body-`[[wikilink]]` agreement described in PROGRAM.md; when a file
violates it, the text report says nothing about drift and every other line reads ✓. `--format
json` shows the findings, so the check works — it just cannot be seen by anyone using the command
the ordinary way.

## Reproduce

Two files in an otherwise empty `PALINODE_DIR`:

`projects/drifter.md` — frontmatter entities that appear nowhere in the body:

```markdown
---
id: projects-drifter
type: ProjectSnapshot
category: projects
entities:
  - project/alpha
  - person/casey
description: A file whose frontmatter entities never appear in the body.
---

This body mentions no wikilinks at all, so both frontmatter entities are drifted.
```

`decisions/pick-a-lock.md` — a Decision with no `priority`:

```markdown
---
id: decisions-pick-a-lock
type: Decision
category: decisions
entities:
  - project/alpha
description: A Decision memory carrying no priority field.
---

We chose the lock. See [[alpha]] for why.
```

Then:

```bash
palinode lint --format json   # wiki_drift: 2 findings; missing_priority: 1 entry
palinode lint                 # neither appears, and no ✓ line mentions them either
```

## Task

Add render blocks for both keys in `palinode/cli/lint.py`, matching the surrounding style exactly
— a `[bold yellow]Heading (count)[/bold yellow]` with one line per finding, or a
`[green]✓ …[/green]` line when empty, since every neighbouring block prints the clean case too.
`wiki_drift` entries are `{"file": str, "warnings": [{"kind": str, "detail": str}]}` — nested like
`source_anchor_issues`, so that block is the closest model. `missing_priority` is a flat list of
paths, like `missing_entities`.

[...]

PR #115, opened 2026-08-11
fix(review): honor TTY-aware output and exit non-zero on API failure (closes #114)

### LINKED ISSUE #114: review: honor TTY-aware output and exit non-zero on API failure
`palinode review` ignores two conventions the rest of the CLI follows. Both are one-line fixes in
`palinode/cli/review.py`.

**1. It is not TTY-aware.** The CLI's contract is human-readable output when interactive, JSON
when piped — `README.md` calls it "TTY-aware output", and `palinode/cli/_format.py` provides
`get_default_format()` for exactly this. Twenty-one CLI modules use it. `review.py:13` hardcodes
`default="text"` instead, so `palinode review | jq` gets a Rich-rendered report rather than JSON.

That is worse than inconvenient: the report goes through `rich.Console`, which hard-wraps at an
80-column non-TTY fallback and consumes `[...]` as style markup. `emit_json`'s docstring in
`_format.py` explains why that path corrupts machine-readable output three separate ways.

**2. It exits 0 when the API call fails.** `review.py:23-25` catches `HTTPStatusError`, prints an
error, and returns — so the process reports success. Fifteen CLI modules exit non-zero on their
error paths. A script that runs `palinode review` in CI cannot tell a clean review from a 500.

## Task

- Use `get_default_format()` for the `--format` default, following any of the twenty-one modules
  that already do (`palinode/cli/status.py` is a short one).
- Make the `HTTPStatusError` path exit non-zero, matching how the other modules do it. The
  `RequestError` path below it is a deliberate local-fallback, not an error — leave it alone.

## Tests

`tests/` — assert the piped default is JSON, and that a failing API response produces a non-zero
exit. Click's `CliRunner` gives you a non-TTY stdout by default, so the first one is a direct
assertion on `result.output` parsing as JSON.

## Sister issue

`palinode/cli/lint.py` has both of these too, and they are covered by #113 along with a separate
rendering bug in that file. The two issues are split by file on purpose — nothing in this PR can
conflict with that one, so both can be worked at the same time.

SkxOverKill:
Closes #114

`palinode review` ignored two CLI conventions the rest of the commands follow. Both are one-line fixes in `palinode/cli/review.py`.

### 1. TTY-aware output
`--format` was hardcoded to `default="text"`, so `palinode review | jq` got a Rich-rendered report instead of JSON. That output goes through `rich.Console`, which hard-wraps at the 80-column non-TTY fallback and consumes `[...]` as style markup (`emit_json`'s docstring in `_format.py` explains the three ways this corrupts machine-readable output).

The default now comes from `get_default_format()` — the same helper 21 other CLI modules use (`status.py`, `archive.py`, …): **text when interactive, JSON when piped**. An explicit `--format` still wins.

### 2. Non-zero exit on API failure
The `HTTPStatusError` path printed an error and `return`ed, so the process reported **success** — a CI script running `palinode review` could not tell a clean review from a 500. It now `raise SystemExit(1)`, matching the other error paths (`archive.py`, `migrate.py`).

The `RequestError` path below it is a deliberate local-in-process fallback (the API being down is not an error), so it's left untouched, as the issue requests.

## Tests

New `tests/test_review_cli.py` (CliRunner gives a non-TTY stdout by default):

- piped default parses as JSON
- explicit `--format json` matches that default
- explicit `--format text` renders the human report (asserts it is *not* JSON)
- an `HTTPStatusError` from the API → non-zero exit + status code printed

`ruff check` passes on both files under the repo's configured ruleset. Related existing suites stay green: `test_review_cli.py` + `test_cli_json_output.py` + `test_review.py` → **19 passed**.

## Sister issue

This is the `review.py` half; #113 covers the same two conventions in `lint.py` plus a rendering bug there. Split by file on purpose — this PR touches nothing #113 touches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/palinode/cli/review.py
+++ b/palinode/cli/review.py
@@ -3,14 +3,14 @@ import click
 from rich.console import Console
 
 from palinode.cli._api import HTTPStatusError, RequestError, api_client
-from palinode.cli._format import emit_json
+from palinode.cli._format import OutputFormat, emit_json, get_default_format
 
 console = Console()
 
 
 @click.command()
 @click.argument("project", required=False)
-@click.option("--format", "fmt", type=click.Choice(["json", "text"]), default="text", help="Output format")
+@click.option("--format", "fmt", type=click.Choice(["json", "text"]), help="Output format")
 def review(project, fmt):
     """Advisory project-memory review.
 
@@ -22,13 +22,14 @@ def review(project, fmt):
         data = api_client.review(project)
     except HTTPStatusError as e:
         console.print(f"[red]Error: API returned {e.response.status_code}[/red]")
-        return
+        raise SystemExit(1)
     except RequestError:
         # Fallback to local in-process review if the API is down.
         from palinode.core.review import run_review
         data = run_review(project=project)
 
-    if fmt == "json":
+    output_fmt = OutputFormat(fmt) if fmt else get_default_format()
+    if output_fmt == OutputFormat.JSON:
         emit_json(data)
         return
 
--- a/tests/test_review_cli.py
+++ b/tests/test_review_cli.py
@@ -1,4 +1,4 @@
-"""CLI-surface tests for ``palinode review`` (issue #114).
+"""CLI-surface tests for ``palinode review`` (https://github.com/phasespace-labs/palinode/issues/114).
 
 ``review`` must follow the two CLI conventions the rest of the commands honour:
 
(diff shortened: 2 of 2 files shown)


Paul-Kyle:
Merged, and thank you — this is a first contribution that reads like a fifth.

You found `get_default_format()` and matched `status.py`'s construction exactly rather than
inventing a second way to do the same thing, and you left the `RequestError` fallback alone and
said why in the module docstring. That distinction — the API being down is not an error, a 500
is — is the part of the issue I'd have expected to get flattened.

Sorry about the wait on CI. First-time fork PRs need a maintainer to release the workflow gate
and I was slow to it; that's on me, not on you. It won't recur on your next one.

PR #125, opened 2026-08-13
fix(parity): annotate the API /timeline alias instead of double-counting it

### LINKED ISSUE #122: parity: API inventory counts the deprecated /timeline alias as its own capability
`GET /timeline/{file_path}` is deprecated in favour of `GET /history/{file_path}?detail=full` —
the handler logs exactly that (`palinode/api/routers/git_history.py`), and the history route's
own docstring describes itself as "formerly the /timeline endpoint". But
`INVENTORY_BACKLOG["api"]` in `palinode/core/parity.py` still lists the two as separate rows, so
the parity inventory counts one capability twice.

This is the same defect that was fixed for the MCP surface and then the CLI surface. **This is
the final timeline surface** — after this, no timeline alias double-counts the history capability.

## Task

In `palinode/core/parity.py`, annotate the canonical API row with its alias instead of carrying a
second row:

```python
"GET /history/{file_path:path}": InventoryBacklogEntry(170, aliases=("GET /timeline/{file_path:path}",)),
```

...and remove the standalone `"GET /timeline/{file_path:path}": 170` entry from the same block.

The endpoint itself must keep working and keep logging its deprecation warning — this changes how
the inventory *counts* it, nothing about runtime behaviour.

Then update `docs/PARITY.md`, whose backlog paragraph currently reads "including the deprecated
MCP `palinode_timeline` and CLI `timeline` aliases". The API alias belongs in that same list.

## Two worked examples to copy

Both are merged, so you can read the whole shape rather than infer it:

- [the MCP fix, public PR #118](https://github.com/phasespace-labs/palinode/pull/118), which
  introduced `InventoryBacklogEntry` and `inventory_backlog_capabilities()`
- [the CLI fix, public PR #121](https://github.com/phasespace-labs/palinode/pull/121), which
  applied it to a second surface — the closest model for this one

## Decisions already made — you do not need to litigate these

- **Annotate, do not delete.** The alias stays live and deprecated; only the double-count goes.
- **Do not promote `history` into `REGISTRY`.** That is separate per-op work tracked under the
  ADR-010 backlog and is out of scope here.
- **Do not touch the route handler.** No behaviour change, no removal of the deprecation log.

## Tests

`tests/test_surface_parity.py` has the pattern twice already — the MCP and CLI alias tests each
assert that the alias is absent from `INVENTORY_BACKLOG`, present in
`inventory_backlog_capabilities()`, and that the capability count is exactly one more than the row
count. Add the API equivalent alongside them.

The three existing inventory guards (`test_no_unregistered_capabilities`,
`test_inventory_accounting_is_not_stale`, `test_inventory_buckets_are_disjoint`) must stay green —
they already route through `inventory_backlog_capabilities()`, so a correct annotation keeps them
passing without modification.

MikeGatsby:
## What this PR does

`GET /timeline/{file_path:path}` is deprecated in favour of
`GET /history/{file_path:path}?detail=full` (the handler already logs this
and the history route's own docstring says as much), but
`INVENTORY_BACKLOG["api"]` still carried both as separate rows, so the
parity inventory counted one capability twice. This is the same defect
already fixed for the MCP surface (#118) and the CLI surface (#121) — this
is the matching fix for the last timeline surface.

Per the issue's own "decisions already made": annotates the canonical
`GET /history/{file_path:path}` row with the alias via
`InventoryBacklogEntry` instead of deleting or promoting anything. No
change to the route handler or its deprecation logging. Updated
`docs/PARITY.md`'s backlog paragraph to list the API alias alongside the
MCP/CLI ones already there.

Closes #122

## Testing

- [x] `python -m pytest tests/test_surface_parity.py -q` — 223 passed,
      1 xfailed (pre-existing, unrelated)
- [x] Added `test_api_inventory_alias_does_not_add_a_capability_row`
      alongside the existing MCP/CLI alias tests, same assertion shape
- [x] The three existing inventory guards
      (`test_no_unregistered_capabilities`,
      `test_inventory_accounting_is_not_stale`,
      `test_inventory_buckets_are_disjoint`) still pass unmodified

### DIFF
--- a/docs/PARITY.md
+++ b/docs/PARITY.md
@@ -136,7 +136,7 @@ Identifier form per surface: MCP = tool name (`palinode_search`); API = `METHOD 
 
 These memory-semantic operations ship on all of MCP/API/CLI today but are not yet promoted into `REGISTRY` with canonical params. They are tracked under #170 (admin/framework surface is in `INVENTORY_INFRA`, not here):
 
-`dedup_suggest`, `diff`, `entities`, `history` (including the deprecated MCP `palinode_timeline` and CLI `timeline` aliases), `ingest`/`ingest-url`, `lint`, `orphan_repair`, `prompt` (list/show/activate), `push`, `session_end`, and the trigger `list`/`remove` + `check-triggers` + `search-associative` API endpoints. `depends/_unblocked` is tracked under #97.
+`dedup_suggest`, `diff`, `entities`, `history` (including the deprecated MCP `palinode_timeline`, CLI `timeline`, and API `GET /timeline/{file_path:path}` aliases), `ingest`/`ingest-url`, `lint`, `orphan_repair`, `prompt` (list/show/activate), `push`, `session_end`, and the trigger `list`/`remove` + `check-triggers` + `search-associative` API endpoints. `depends/_unblocked` is tracked under #97.
 
 Promoting each (registry `Operation` + canonical params + removing its backlog entry) is the per-op work the issue tracks; this contract makes the gap explicit and prevents *new* unregistered ops from slipping in alongside them.
 
--- a/palinode/core/parity.py
+++ b/palinode/core/parity.py
@@ -675,10 +675,11 @@ INVENTORY_BACKLOG: dict[Surface, dict[str, int | InventoryBacklogEntry]] = {
         "GET /diff": 170,
         "GET /entities": 170,
         "GET /entities/{entity_ref:path}": 170,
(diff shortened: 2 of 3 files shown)


MikeGatsby:
Closing — didn't see #123 before pushing this (my mistake, should have checked for open PRs against this issue before submitting). Looks like it covers the same fix plus a changelog entry this PR is missing. Thanks for the good work, deferring to that one.

PR #132, opened 2026-08-15
Do not overwrite an unrelated memory on a derived-slug collision

### LINKED ISSUE #129: save silently overwrites an existing memory when the slug is derived from content
When `slug` is omitted, the save path derives one from the opening words of the content. Two saves
whose openings agree resolve to the same file, and **the second overwrites the first** — with no
error, no warning, and a success receipt identical to a fresh create.

Reproduced on current `main` with three saves that share an opening line and differ afterwards:

```
save 0: rel_path=insights/user-thanks-that-helps-a-lot.md
save 1: rel_path=insights/user-thanks-that-helps-a-lot.md
save 2: rel_path=insights/user-thanks-that-helps-a-lot.md
distinct rel_paths: 1
files on disk: ['user-thanks-that-helps-a-lot.md']
surviving file mentions  ALPHA: False | BRAVO: False | CHARLIE: True
```

Two memories are gone from the queryable store. They are recoverable from git history — three
commits to one path — but nothing told the caller anything happened. Exposure scales with write
volume and with how formulaic the content is: human-authored saves rarely collide, while anything
saving conversational turns, logs, or telemetry collides constantly.

**There is a precedent in the codebase, but do not copy it.**
`palinode/ingest/pipeline.py:311-314` attempts this job for research ingestion:

```python
if os.path.exists(filepath):
    slug += f"-{stable_md5_hexdigest(content[:100])[:6]}"
    filename = f"{today}-{slug}.md"
    filepath = os.path.join(config.palinode_dir, "research", filename)
```

**It has two defects, and copying it verbatim would reproduce the bug you are fixing:**

1. **It hashes only `content[:100]`.** Two memories that agree for their first 100 characters and
   differ afterwards hash *identically*, so they land on the same disambiguated filename and the
   second still overwrites the first. Confirmed — with two bodies sharing a 120-character prefix,
   both produce the suffix `13365e`.
2. **It checks existence once.** If the disambiguated name is also taken, it overwrites without
   looking.

**What to change** in `palinode/core/save.py`, for the derived-slug path only:

- Disambiguate using a hash of the **full content**, not a prefix — or loop until the candidate
  filename is unused (`-2`, `-3`, …). Either is fine; a single prefix-hash check is not.
- If the existing file's content is byte-identical, keeping the current overwrite is correct — that
  is a re-save of the same memory, not a collision.

**Tests to include:**

- Two saves sharing an opening line, differing afterwards → two distinct files, both retrievable.
- **Two saves whose content is identical for the first 100 characters and differs only after
  character 100 → still two distinct files.** This is the case a prefix hash silently fails, so it is
  the test that proves the fix rather than the pattern.

**Two things to leave alone, both deliberate:**

1. **An explicitly-passed `slug` must keep overwriting.** That is the documented escape hatch and
   callers rely on it for idempotent updates. Only the derived case should change.
2. **Do not extract a shared helper into `palinode/ingest/pipeline.py`, or fix its version here.**
   Keep your change inside `palinode/core/save.py`. There is another issue open against that file's
   docstrings and we would rather not make either of you rebase.

`update_policy` is not the lever here, in case it looks like one: it is sticky frontmatter recording
write semantics, not a branch that decides whether the file gets replaced.

kratos0718:
Fixes #129.

## What was happening

When `slug` is omitted, it is derived from the opening words of the content. Two saves whose openings agree resolve to the same path, and the second replaces the first — no error, no warning, and a receipt identical to a fresh create. Reproducing your three-save case in a test:

```
expected 3 distinct paths, got
  ["insights/user-thanks-that-helps-a-lot.md",
   "insights/user-thanks-that-helps-a-lot.md",
   "insights/user-thanks-that-helps-a-lot.md"]
assert 1 == 3
```

## The distinction the fix turns on

The existing-path branch is deliberate, and I did not want to disturb it. Re-saving an **explicit** slug is an update of the same logical memory — that is the case `created_at` carry-forward exists for, and it should keep overwriting.

A **derived** slug colliding is a different thing: two unrelated memories that happen to start alike. So the save path now records whether the slug was derived, and only that case is disambiguated. Explicit-slug behaviour is byte-for-byte unchanged.

## Behaviour

- Derived slug, path free → unchanged.
- Derived slug, path taken by **different** content → next free `slug-2`, `slug-3`, … Nothing is overwritten and the receipt reports the path actually written.
- Derived slug, path taken by **identical** content → same file, no duplicate. The incoming content hash is compared against the `content_hash` already in the target's frontmatter, so the same memory arriving twice does not litter the store.
- Explicit slug → unchanged, still overwrites.

The suffix search is bounded, with a timestamp fallback so it always terminates.

## Open question

I kept explicit-slug collisions silent, since that is the current contract. If you would rather they became loud too — a distinct receipt field, or a rejection when `update_policy` does not permit replacement — I am happy to follow up; it felt like your design call rather than mine, so I did not fold it into a bug fix.

## Tests

`tests/test_save_derived_slug_collision.py`:

| test | before | after |
|---|---|---|
| `test_colliding_derived_slugs_do_not_overwrite` | FAIL — 1 distinct path, ALPHA and BRAVO gone | pass |
| `test_identical_content_is_not_duplicated` | FAIL | pass |
| `test_explicit_slug_still_overwrites` | pass | pass |

The third passes before and after on purpose — it pins the behaviour this change must *not* alter.

```
pytest -m "not slow"   3114 passed, 10 skipped, 5 xfailed
ruff check             All checks passed
bandit -r palinode -ll clean
```

### DIFF
--- a/palinode/core/save.py
+++ b/palinode/core/save.py
@@ -159,6 +159,60 @@ def _normalize_claims_or_reject(raw: Any, memory_ref: str) -> list[dict[str, Any
         raise SaveValidationError(str(exc))
 
 
+def _disambiguate_derived_slug(
+    slug: str, file_path: str, content: str
+) -> tuple[str, str]:
+    """Give a content-derived slug its own path when one is already taken.
+
+    Two saves whose opening lines agree derive the same slug, and writing both
+    to that path leaves only the last one in the store -- no error, no warning,
+    and a receipt indistinguishable from a fresh create. The earlier memories
+    survive only in git history, which the queryable store never consults.
+
+    Re-saving *identical* content is left alone: that is the same memory
+    arriving twice, and suffixing it would litter the store with duplicates.
+
+    Only ever called for derived slugs. An explicit slug that collides is an
+    update of the same logical memory and keeps its overwrite semantics.
+    """
+    if not os.path.exists(file_path):
+        return slug, file_path
+
+    # Same content arriving twice is the same memory, not a collision. Compare
+    # against the content_hash already recorded in the file's frontmatter
+    # rather than re-deriving a body: parse_markdown returns chunks, not text.
+    from palinode.core import parser as _parser
+
+    incoming_hash = hashlib.sha256(content.encode()).hexdigest()
+    try:
+        with open(file_path, "r") as existing:
+            existing_meta, _ = _parser.parse_markdown(existing.read())
+        if str(existing_meta.get("content_hash", "")) == incoming_hash:
+            return slug, file_path
+    except (OSError, ValueError, yaml.YAMLError):
+        # Unreadable or unparseable: suffix rather than overwrite something
+        # we could not inspect.
+        pass
+
+    directory = os.path.dirname(file_path)
+    # Bounded so a pathological directory cannot spin here; the timestamp
+    # fallback below always terminates.
+    for suffix in range(2, 1000):
+        candidate_slug = f"{slug}-{suffix}"
+        candidate_path = os.path.join(directory, f"{candidate_slug}.md")
+        if not os.path.exists(candidate_path):
+            logger.info(
+                "derived slug %r already taken; saving as %r to avoid "
+                "overwriting an unrelated memory",
+                slug,
+                candidate_slug,
+            )
+            return candidate_slug, candidate_path
+
+    candidate_slug = f"{slug}-{int(time.time() * 1000)}"
+    return candidate_slug, os.path.join(directory, f"{candidate_slug}.md")
+
+
 def save_memory(
     *,
     content: str,
@@ -262,7 +316,13 @@ def save_memory(
         # Prevent any potential JSON escape or traversal exploits if user defines slug
         slug = re.sub(r'[^a-z0-9]+', '-', slug.lower()).strip('-')
 
+    # Whether the slug was chosen by the caller or inferred from the content.
+    # An explicit slug that collides is an update of the same logical memory and
+    # must keep overwriting. A *derived* slug that collides is an accident
+    # between two unrelated memories, and overwriting there loses data silently.
+    slug_was_derived = False
     if not slug:
+        slug_was_derived = True
         slug = re.sub(r'[^a-z0-9]+', '-', content.split('\n')[0].lower()[:30]).strip('-')
         if not slug:
             slug = str(int(time.time()))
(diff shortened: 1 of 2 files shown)


Paul-Kyle:
The main collision fix is right, but there is one idempotency case the current tests don't reach.

If A owns `slug.md`, the first save of B correctly lands at `slug-2.md`. Re-saving B then compares
only against A at the base path, skips `slug-2.md` because it exists without checking its recorded
`content_hash`, and writes the same B content again at `slug-3.md`.

Please make the suffix scan reuse an existing candidate whose `content_hash` matches the incoming
content, and add this regression:

1. save A → `slug.md`
2. save B → `slug-2.md`
3. save B again → still `slug-2.md`, with exactly two files in the directory

Everything else in the approach looks good, including keeping explicit-slug overwrites and failing
away from an unreadable file rather than clobbering it. Once this case is green, I'll re-review the
exact head for merge.

Paul-Kyle:
One addition to the note above, and it's our sequencing rather than anything on your side: #131
merged a couple of hours ago, so this branch now also shows as behind `main`.

So when you push the idempotency fix, update onto `main` in the same go — GitHub's *Update branch*
button will do it. Flagging it now so you don't push, watch the checks go green, and then find the
merge still blocked.

Paul-Kyle:
Checking in rather than chasing — no deadline attached to this.

The idempotency case and the rebase are both still the only things between this and a merge, and
your implementation is otherwise exactly what the issue wanted. If you're still up for it, push when
you get a chance and I'll merge on green.

If you'd rather hand it back, that's completely fine and no explanation needed — say the word and
I'll take it from here with your authorship preserved on the commit. Either way the work you did
stands.

kratos0718:
Pushed. Good catch, that was a real hole.

The scan skipped any occupied candidate without looking at what was in it, so a memory that lost the base path on its first save landed on `-2` and then climbed to `-3`, `-4` on every later save of the same content. The base-path hash check is now a small helper (`_holds_this_content`) and the suffix scan calls it too, so an occupied candidate holding this exact content gets reused instead of stepped over.

Regression test is `test_resaving_a_suffixed_memory_reuses_its_own_file`, doing your three steps and asserting the directory ends up with exactly the two expected files. I checked it fails on the previous head, landing on `-3.md` as you described. Full suite is 3119 passed, 10 skipped, 5 xfailed, and ruff is clean.

Branch is already on top of `main`, so it should be showing green now.

PR #140, opened 2026-08-15
docs: rewrite generated docstrings in pipeline.py

### LINKED ISSUE #136: Finish the generated-docstring cleanup in ingest/pipeline.py
**If you'd like to take this, please comment here first and wait to be assigned before opening a
PR.** Two contributors landed on the same issue yesterday because we had no such line and no
assignment; that was our fault, and this is the fix.

#130 asked for five specific docstring lines in `palinode/ingest/pipeline.py` to be rewritten and
#131 did exactly that. But five lines was the wrong scope: the file has roughly twenty-five lines in
the same generated register, and because #130 named individual lines rather than docstrings, two of
them are now **half-converted** — plain English beside word-salad inside the same docstring. That is
my error, not #131's.

**Still to rewrite, all in `palinode/ingest/pipeline.py`:**

| Line | Current text |
|---|---|
| 60 | `Scan ingestion inboxes iteratively routing newly seeded payloads safely.` |
| 85 | `Invokes explicit parsing algorithms depending heavily on disk extensions.` |
| 88 | `filepath (str): Evaluated system path triggering event cycles.` |
| 91–92 | `Emits absolute resulting saved context path safely if correctly triggered. None otherwise explicitly failing or skipping safely.` |
| 102 | inline comment `# extract audio natively through transcriptor boundaries` |
| 139 | inline comment — **trailing clause only**: `# Cap for very large PDFs using explicit configuration logic constraints`. "Cap for very large PDFs" is accurate; drop or replace what follows |
| 113 | `Extract semantic text blocks from unparsed PDF layouts formatting as markdown references.` |
| 116 | `filepath (str): Absolute raw OS path string matching input PDF schema formats.` |
| 154 | `Streams encoded chunks securely into configured remote transcriptor GPU pipelines.` |
| 157 | `filepath (str): Target physical file footprint mapping active media formats string sequences.` |
| 158 | `name (str): Native system filename.` |
| 161 | `str | None: The resulting context pathway saving to memory if perfectly transcribed.` |
| 193 | `Evaluates raw strings determining if parsing redirects logic to URL processors explicitly.` |
| 196 | `filepath (str): Evaluated system path triggering event cycles.` |
| 197 | `name (str): Root file target name.` |
| 200 | `str | None: Active reference resulting memory path.` |
| 218 | `Reads MacOS .webloc XML files or explicit Windows .url links cleanly fetching targets.` |
| 221–222 | `Standard platform shortcut reference footprint mapping system shortcut paths.` / `Link target OS base payload schema string sequence names.` |
| 225 | `Processed semantic node markdown resulting absolute DB disk array links.` |
| 245–246 | `Downloads HTTP response targets applying readability algorithms scrubbing out semantic fat strings locally.` |
| 264 | inline comment — **trailing clause only**: `# Simple readability: strip HTML tags cleanly securing memory`. "Simple readability: strip HTML tags" is accurate; drop or replace what follows |
| 250 | `name (str): Fallback generic slug text explicitly formatting the result paths correctly.` |
| 253 | `str | None: Successful parsed response local filepath.` |
| 293–294 | `Generates a perfectly modeled Frontmatter YAML formatted chunk safely storing contexts across DB indexes.` |
| 299 | `source_file (str): System path tracking originating sources schema tracking logs.` |
| 301 | `file_type (str): Explicit mapping array categories types string.` |

The last two blocks are the half-converted ones: #131 correctly rewrote `url`, `name`, `content`,
`source_url` and the `Returns:` line in those docstrings, so please match the register it
established rather than inventing a third style.

**A useful reference point for the `name` parameter:** `ingest_pdf` already documents the same thing
correctly at `:117` — `name (str): Document basename.` Lines 158 and 197 are the generated variants
of that one line, which is why they are on the list rather than left as terse-but-fine.

**Leave alone:** the module docstring (1–10), `is_safe_url` (35), and the comments at 124
(`# pymupdf`), 230, 235, and 324–325 — all accurate as written. Only the trailing clauses of the
comments at 139 and 264 need touching; their opening halves are correct and should survive.

**Scope:** docstrings and those three inline comments, this file only. **No behaviour changes, no new
functions, no deletions** — the diff should be docstring lines and nothing else.
`palinode/indexer/watcher.py` has the same problem and is filed separately; please leave it to
whoever takes that one.

JodVarun:
Fixes #136 

Rewrote machine-generated docstrings and inline comments in palinode/ingest/pipeline.py to natural English, matching the style established in #131 per issue #130.

### DIFF
--- a/palinode/ingest/pipeline.py
+++ b/palinode/ingest/pipeline.py
@@ -57,7 +57,7 @@ def is_safe_url(url: str) -> bool:
 
 
 def process_inbox() -> None:
-    """Scan ingestion inboxes iteratively routing newly seeded payloads safely."""
+    """Scan ingestion directories for new files."""
     raw_dir = os.path.join(config.palinode_dir, config.ingestion.inbox_dir)
     processed_dir = os.path.join(config.palinode_dir, config.ingestion.processed_dir)
     os.makedirs(raw_dir, exist_ok=True)
@@ -82,14 +82,13 @@ def process_inbox() -> None:
 
 
 def process_file(filepath: str) -> str | None:
-    """Invokes explicit parsing algorithms depending heavily on disk extensions.
+    """Invoke parser based on file extension.
 
     Args:
-        filepath (str): Evaluated system path triggering event cycles.
+        filepath (str): Path to the input file.
 
     Returns:
-        str | None: Emits absolute resulting saved context path safely if correctly triggered.
-            None otherwise explicitly failing or skipping safely.
+        str | None: Returns absolute path of the saved file if successful, or None.
     """
     ext = Path(filepath).suffix.lower()
     name = Path(filepath).stem
@@ -99,7 +98,7 @@ def process_file(filepath: str) -> str | None:
     elif ext in (".m4a", ".mp3", ".wav", ".ogg", ".flac"):
         return ingest_audio(filepath, name)
     elif ext in (".mp4", ".mkv", ".mov", ".webm"):
-        return ingest_audio(filepath, name)  # extract audio natively through transcriptor boundaries
+        return ingest_audio(filepath, name) # Extract audio track
     elif ext in (".md", ".txt"):
         return ingest_text(filepath, name)
     elif ext in (".url", ".webloc"):
@@ -110,10 +109,10 @@ def process_file(filepath: str) -> str | None:
 
 
 def ingest_pdf(filepath: str, name: str) -> str | None:
-    """Extract semantic text blocks from unparsed PDF layouts formatting as markdown references.
+    """Extract text blocks from unparsed PDF layouts formatted as Markdown.
 
     Args:
-        filepath (str): Absolute raw OS path string matching input PDF schema formats.
+        filepath (str): Path to input PDF.
         name (str): Document basename.
 
     Returns:
@@ -136,7 +135,7 @@ def ingest_pdf(filepath: str, name: str) -> str | None:
             logger.warning(f"Empty PDF: {filepath}")
             return None
 
-        # Cap for very large PDFs using explicit configuration logic constraints
+        # Cap for very large PDFs
         capped_len = config.ingestion.pdf_max_chars
         
         return write_research_file(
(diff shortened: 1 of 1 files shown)


Paul-Kyle:
Close — the comments are handled exactly right, and I want to flag that first because it is the
instruction most people skim: you kept "Cap for very large PDFs" and "Simple readability: strip HTML
tags" and dropped only the generated tails, which is what the issue asked for.

Two things before I merge.

**Three lines from the issue list are still untouched**, all `Returns:` lines — current 160, 199 and
251:

- `str | None: The resulting context pathway saving to memory if perfectly transcribed.`
- `str | None: Active reference resulting memory path.`
- `str | None: Successful parsed response local filepath.`

**And four of the rewritten lines don't match what the code does.** I would rather raise these than
merge them, because the whole point of the issue is that a confident wrong docstring costs a reader
more than an obviously empty one:

- **`ingest_url_file` Returns (`:224`)** — "Returns target URL extracted from shortcut file." It
  returns `ingest_url(...)`, so the value is the saved research-file path, or `None`. This is the one
  I would most like fixed: it is new text and it is confidently wrong.
- **`ingest_audio` (`:153`)** — "Stream audio chunks." It opens the file and posts it once as
  multipart; nothing is chunked.
- **`ingest_text` (`:192`)** — "Determine if the input string is a URL." That is the branch test, not
  the function: it ingests the text, handing off to URL ingestion when the content looks like a URL.
- **`write_research_file` (`:291`)** — "Generate YAML frontmatter." It writes the research file and
  returns its path; the frontmatter is a step inside that.

No rush, and none of this is a rewrite — it is seven lines. Push when you have them and I will merge
on green.

JodVarun:
Thanks for the detailed feedback! I've updated the 3 untouched Returns: lines and corrected the 4 docstrings to accurately reflect function behavior. Pushed the update!

Paul-Kyle:
All seven addressed — merging. Apologies for the overnight wait on the corrected push.

The one I most wanted right is `ingest_url_file`'s return: it now says the saved research-file path
rather than the extracted URL, which is what `return ingest_url(...)` actually gives you. A wrong
docstring costs a reader more than a meaningless one, so that correction is the difference between
this file being fixed and being differently broken.

This is your first contribution here and it took a review round to land — that is completely normal,
and the second pass was clean. Thanks for the patience.

PR #141, opened 2026-08-16
fix: preserve embedding outage status for trigger endpoints

### LINKED ISSUE #137: Trigger endpoints return a generic 500 when the embedder is unavailable
**If you'd like to take this, please comment here first and wait to be assigned before opening a
PR.** Two contributors landed on the same issue yesterday because we had no such line and no
assignment; that was our fault, and this is the fix.

When the embedding backend is down, the same failure produces different answers depending on which
endpoint you hit.

`POST /search` and its neighbours return **503** with a message naming the backend, the model, and a
`palinode doctor` recovery hint — `palinode/api/routers/search.py` catches
`embedder.EmbeddingUnavailable` and routes it through `_embedding_unavailable_503` (`:35`) at four
call sites.

`POST /triggers` and `POST /check-triggers` return a generic **500**. Both call `embedder.embed()`
(`palinode/api/routers/triggers.py:33` and `:67`) under a broad `except Exception` that turns the
typed exception into `_safe_500`, so the operator sees "Trigger creation failed" and none of the
information the exception exists to carry. A 500 also implies "we broke" where a 503 correctly says
"a dependency is down, retry."

**What to change:** catch `embedder.EmbeddingUnavailable` in both trigger handlers *before* the broad
`except Exception`, and return the same typed 503 the search router does.

`_embedding_unavailable_503` is currently module-private to `search.py`. Preferred approach: move it
to `palinode/api/_util.py` and import it from both routers — a small mechanical change, and no other
work is in flight in `search.py` so touching it is safe. Duplicating a small equivalent in
`triggers.py` is acceptable if you'd rather keep the diff to one file; say which you chose and why.

**Please add a regression test for each endpoint separately — one for `POST /triggers` and one for
`POST /check-triggers`** — asserting 503 rather than 500 when the embedder raises. They are two
independent handlers with their own `try` blocks, so a single test covers only the one it calls, and
the untested handler is exactly where this would come back.

**One thing you'll notice:** both handlers have an `if not emb:` check just after the call. That
dates from an older contract where `embed()` returned `[]` on failure; it now raises instead, so
those branches are unreachable. Removing them is welcome but optional — say so in the PR either way
so the reviewer isn't guessing.

JulianZJN:
Closes #137

## Summary

- Move the shared embedding-unavailable 503 helper into `palinode/api/_util.py` and reuse it from the search and trigger routers.
- Catch `EmbeddingUnavailable` before the generic 500 handler in `/triggers` and `/check-triggers`.
- Leave the existing empty-embedding branches unchanged and add one regression test per endpoint.

## Tests

- `uv run pytest tests/test_search_embedder_outage.py tests/test_triggers_vec_upsert.py -q` (15 passed)
- `uv run ruff check palinode/ tests/ scripts/`
- `uv run bandit -r palinode/ -ll` (no issues identified)
- `git diff --check`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,9 @@ All notable changes to Palinode. Format follows [Keep a Changelog](https://keepa
 
 ### Fixed
 
+- Preserve the embedding-backend 503 response for `/triggers` and
+  `/check-triggers` instead of converting outages into generic 500 errors.
+
 ### Removed
 
 ### Security
--- a/palinode/api/_util.py
+++ b/palinode/api/_util.py
@@ -19,6 +19,7 @@ from typing import Any
 
 from fastapi import HTTPException
 
+from palinode.core import embedder
 from palinode.core.config import config
 from palinode.core.retrieval_log import RetrievalLogger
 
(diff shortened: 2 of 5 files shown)


Paul-Kyle:
Merging — and sorry this sat overnight; the delay was mine, not anything about the PR.

You did all four things you said you would, including the part that is easiest to half-do: two
separate regression tests, one per endpoint, rather than one test covering the handler you happened
to touch first. That is exactly why the issue asked for them that way.

Moving the helper into `_util.py` rather than duplicating it was the right call too — `/triggers` and
`/check-triggers` now fail the same way `/search` does, which is the whole point of the issue.

PR #145, opened 2026-08-20
fix(lint): render wiki drift and missing priority in text output

### LINKED ISSUE #113: palinode lint silently drops two checks from its default output

sidsri14:
## Summary

Fixes the two checks that `palinode lint`'s default **text** output silently
dropped even though they run and are surfaced in `--format json`:

- **Wiki contract drift** (`wiki_drift`)
- **Missing human priority** (`missing_priority`)

The text report now renders both, including an explicit clean-state line for
each ("No wiki drift", "All core and Decision memories have priority"), and
additionally reports the scanned-file total. A coverage guard in
`tests/test_lint.py` asserts that every key returned by `run_lint_pass()` has a
text representation, so a future check can never be silently dropped again.

## Why

The issue's title ("silently drops") is literal: the deterministic checks run
every lint pass and appear in JSON output, but the human-readable report never
showed them. A user running `palinode lint` with no flags saw no sign that the
checks existed — no finding, and no reassurance when the corpus was clean.
Clean-state lines matter as much as findings here: a missing-but-clean check is
indistinguishable from a check that no longer runs.

## Root cause

`palinode/cli/lint.py` renders each check into the text report, but the
`wiki_drift` and `missing_priority` buckets had no text branch at all. The
underlying checks in `palinode/core/lint.py` were already producing the data.

## Behaviour change

- `palinode lint` (text) now lists wiki-drift warnings and missing-priority
  files when present, and prints a green clean-state line when not.
- Text output now includes `Files scanned: N`.
- No change to JSON output or to the checks themselves.

## Tests

- `tests/test_lint.py`: five new tests —
  - report rendering for wiki drift (finding present)
  - clean-state contract for wiki drift
  - report rendering for missing priority (finding present)
  - clean-state contract for missing priority
  - **coverage guard**: every `run_lint_pass()` result key has a text marker
    that actually appears in the output
- Verified the new tests **fail before** the renderer change and **pass after**
  (TDD).
- `tests/test_changelog_structure.py` passes.
- Local checks all pass: `ruff check palinode/ tests/ scripts/`, `bandit -r
  palinode/ -ll` (no findings at or above medium), `git diff --check`, and the
  repository's custom CI linters (httpx-monopoly, write-choke-point) with their
  test suites.

## Limitations (baseline failures, reproduced on untouched upstream)

Running the full suite locally on this machine (Windows) yields 196 failures +
5 errors across the repo (2819 passed). These are **pre-existing baseline
failures, not caused by this change**:

- A full-suite run on a clean checkout of upstream `main` (`eb4e0de`) produces
  **208 failures + 7 errors** — i.e. the same or more. Every lint-relevant
  failure (6 tests) reproduces identically on clean upstream.
- The dominant cause is a path-separator mismatch: tests assert forward-slash
  paths (`daily/file.md`, `decisions/claim.md`) while `os.path.relpath` on
  Windows returns backslashes (`daily\file.md`). This includes the two
  `tests/test_lint.py` baseline failures
  (`test_lint_relative_dates_exempts_daily_logs`,
  `test_lint_missing_fields_numerator_matches_total_files_denominator`), which
  fail on clean upstream too.
- A handful of MCP/config/hook tests are order-dependent in a full run: 5 tests
  that failed during the branch's full run pass in isolation (and pass on the
  baseline run), so they are flaky rather than regressions.
- The 5 errors are the Obsidian integration tests, which require the Obsidian
  vault demo harness and are excluded from CI's unit-test run
  (`pytest tests/ --ignore=tests/integration`).

CI runs `pytest tests/ --ignore=tests/integration` on Linux; the lint-relevant
tests in this PR pass.

## Notes

- CHANGELOG entry added under `## Unreleased > ### Fixed` per the contribution
  rules.
- Co-authored with Claude (Anthropic), which performed the implementation,
  test writing, and verification per the repository's AI-assistance disclosure
  expectations.

Fixes #113


Paul-Kyle:
Thanks for this, and apologies for the silence — you opened it two days ago and should have had a
reply the same day.

Reviewing now. It's green and it's the right shape at a glance: both dropped checks rendered in the
text path, with the test covering them. I'll come back with anything substantive; otherwise this
merges shortly.

Issue #150, opened 2026-08-26
fix(search): stop presenting fused rank as match confidence


Paul-Kyle:
Hybrid search reports a fused score of `1.0` for candidates whose underlying cosine similarity is
barely above the floor, so a caller cannot distinguish a strong match from a marginal one.

Measured on `a27d9b3` (v0.14.0) with real BGE-M3 embeddings and the real sqlite-vec + FTS5 store, by
@WilliamK112 while investigating abstention: across 32 searches that returned a hit for a query
whose answer was deliberately absent from the corpus, **every one reported a top fused score of
1.0**, while the corresponding raw cosine ranged **0.402–0.459** (median 0.421).

This is separable from the threshold question. Whatever floor the project settles on, RRF
normalisation is rank-based — the top-ranked item scores 1.0 whether it is an excellent match or the
least-bad of a weak field. A caller reading the fused score as confidence is reading something that
does not carry that information.

The REST `/search` payload already exposes both `score` (the post-fusion rank value) and
`raw_score` (vector cosine when available), and prompt injection deliberately renders `raw_score`.
The remaining defect is presentation: MCP formats `score` as "% match", the human CLI's
`--score` prints it, and the generic API key invites callers to read it as relevance confidence.
Align the naming and rendering across surfaces: use `raw_score` where available, label the fused
value as rank/fusion, and define the BM25-only case where cosine is unavailable. Exact compatibility
shape is open.

Not blocking anything; it has presumably behaved this way since hybrid search landed. Filing it
separately so it doesn't wait on the abstention protocol work.

chiruu12:
Happy to take this. One thing to agree before I write it, since the compatibility shape is open.

What I would do: keep `score` on the REST payload as it is, and change only how the surfaces read it. MCP stops formatting `score` as "% match" and shows `raw_score` as the similarity, with the fused value labelled as rank. The CLI `--score` does the same. Where cosine is unavailable, BM25-only, nothing claims a similarity at all rather than showing a rank dressed as one.

That keeps the payload backward compatible and moves only the presentation, which is where you put the remaining defect. If you would rather rename the REST field and take the break, say so and I will do that instead.

Which way do you want it?

Paul-Kyle:
Presentation-only. Keep `score` and `raw_score` on the REST payload exactly as they are — your reading of the issue is right, the defect is in what the surfaces claim the number means, not in the number being absent.

[...]

PR #171, opened 2026-08-31
fix(search): show similarity, not fused rank, as match

chiruu12:
Part of #150, taking the two Python surfaces. The OpenClaw and shared-core TypeScript renderers and the hook pair are not in this PR.

Presentation only, as you asked. `score` and `raw_score` on the REST payload are untouched.

The three cases are decided in one place, `palinode/core/scoring.py`, rather than once per surface. You flagged the duplicated jq as a wart you did not want extended; two Python renderers making the same judgement separately is the same wart in a different language, so MCP and the CLI now both call `describe_match`.

| Result | Before | After |
|---|---|---|
| cosine 0.421, fused 1.0 | `100% match` | `42% match` |
| `raw_score: null` (BM25-only) | `100% match` | `keyword match, rank 1.00` |
| `raw_score` absent (pre-0.12) | `100% match` | `rank 1.00` |

I kept the null and absent cases apart, as you asked. A null means the ranker told us this hit had no vector arm, so the surface says which arm did fire. An absent field means the server never sent one and the arm is unknown, so the rank is all it says. Neither claims a similarity.

On the naming: `rank`, used everywhere. For the BM25-only case I went with a word rather than silence. `keyword match` says what actually matched, and dropping the number entirely loses the ordering a reader is using the flag to see. It reads as a match without ever reading as a similarity, which was the property you wanted.

**Validation**

- `pytest -q`: 3263 passed, 10 skipped, 5 xfailed
- `ruff check`: clean
- The three surface tests fail on unfixed renderers and pass on the fix. Reverting just `mcp.py` and `search.py` gives `[1.00] notes/a.md` from the CLI and `(100% match)` from MCP.

The tests use the measured numbers from the issue: 0.421 as the median of the 32 searches, and each of the three points across the 0.402 to 0.459 range asserted to render under half.

### DIFF
--- a/palinode/cli/search.py
+++ b/palinode/cli/search.py
@@ -4,6 +4,7 @@ from palinode.cli._api import HTTPStatusError, api_client
 from palinode.cli._format import print_result, console, OutputFormat, get_default_format
 from palinode.core.config import config
 from palinode.core.parity import CATEGORIES, MEMORY_TYPES
+from palinode.core.scoring import describe_match
 
 
 def _cli_resolve_context() -> list[str] | None:
@@ -125,7 +126,7 @@ def search(
                 return
             
             for res in results:
-                score_str = f"[{res['score']:.2f}] " if score else ""
+                score_str = f"[{describe_match(res)}] " if score else ""
                 title = res.get("file", "Untitled")
                 # prefer the API-provided snippet — already match-windowed
                 # and bounded. Fall back to the legacy blind-truncation path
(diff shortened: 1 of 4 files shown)


Paul-Kyle (approved):
Merging. The shared helper is the right shape here: MCP and CLI make the same three-way decision once, numeric raw cosine is the only value presented as match percentage, present-null is labelled as keyword-only plus rank, and an absent legacy field is rank-only. I re-read the refreshed four-file diff at `5d591f3` after the branch update and all 9 required checks are green. This remains correctly partial, so #150 stays open for the TypeScript and hook renderers.

PR #209, opened 2026-09-13
feat: support authenticated OpenAI-compatible embedding endpoints

### LINKED ISSUE #203: embeddings: support bearer-authenticated OpenAI-compatible endpoints with configurable paths
`embeddings.primary.dialect: "openai"` (added so palinode can embed without Ollama) posts to an OpenAI-compatible embeddings endpoint. The README demonstrates local, unauthenticated llama.cpp, vLLM, and LM Studio deployments rooted at a bare host; each server can also be configured to require an API token.

**There are two gaps across the hosted-provider class.** OpenAI and Voyage need the first; Gemini needs both, and either one alone is insufficient for that provider. Both gaps were checked against `main`:

**1. Nothing on the path can send a credential.**

- `PrimaryEmbeddingConfig` (`palinode/core/config.py`, lines 91–113) carries `model`, `url`, `dimensions`, `timeout_seconds`, `connect_timeout_seconds`, `dialect` — and no key.
- `OllamaClient.__init__` constructs `httpx.Client()` with no default headers (`palinode/core/ollama_client.py`, line 556).
- The request site is `self._client.post(url, json=payload, timeout=timeout)` — no `headers` argument on the path at all.
- There is no `api_key`, `authorization`, or `bearer` anywhere in `config.py`, `ollama_client.py`, or `embedder.py`.

**2. The request path is hard-coded and only half-normalized.** `_OPENAI_EMBED_PATH = "/v1/embeddings"` (`ollama_client.py` line 347) is appended to a base that is trimmed only when it already ends in `/v1` (`_openai_embed_base_url`, line 350). Providers do not agree on that root. Google's OpenAI-compatible surface is rooted at `…/v1beta/openai/` and its embeddings endpoint is `…/v1beta/openai/embeddings`, so:

```
configured:  https://generativelanguage.googleapis.com/v1beta/openai
palinode →   https://generativelanguage.googleapis.com/v1beta/openai/v1/embeddings   ← 404
Google is →  https://generativelanguage.googleapis.com/v1beta/openai/embeddings
```

A credential alone still yields a `404`; a correct path alone still yields a `401`.

**Neither failure is silent, to be fair to the current code.** Every `4xx` is treated as a permanent caller error — no retry, `trip=False` so the circuit breaker stays closed, and the server's own response body is carried into the raised error (`ollama_client.py` lines 674–700). The diagnostics are fine. The capability is missing.

**There is a workaround, and naming it is the honest way to argue for the fix.** Point `url` at a local proxy that speaks the OpenAI embeddings shape inbound, holds the key, rewrites the path, and forwards upstream. That is not hypothetical — it is the pattern this repo already documents for the *chat* role, where the example fallback ladder ends at a "local Sonnet shim" on loopback (`palinode.config.yaml.example` line 115). It works, and it is why nobody has hit this.

It is still the wrong default. A sidecar process, its own supervision, and its own failure mode are a lot of machinery to add a header and fix a suffix, and the shim has to be written by each operator — no shim ships with palinode. The circuit breaker, retry policy, and per-input error handling that make the embed path reliable all live on *this* side of the proxy, so a shim outage looks like an embedding outage with none of that diagnosis.

**Why this is worth doing generically rather than as a provider integration.** OpenAI and Voyage serve the `/v1/embeddings` shape directly; Google serves the same request and response contract at its own root ([docs](https://ai.google.dev/gemini-api/docs/openai)). So with a credential and an addressable path, a hosted **text** embedding provider becomes configuration:

```yaml
embeddings:
  primary:
    dialect: "openai"
    url: "https://generativelanguage.googleapis.com/v1beta/openai"
    # plus whatever the chosen shape is for the endpoint path — see below
    model: "gemini-embedding-001"
```

Palinode stays local-first by default: none of this changes anything unless an operator sets a key.

**Deliberately not claimed:** this does not bring Cohere in — it serves `/v2/embed` with a different request and response contract, and would need a real adapter. It also does not unlock Gemini *multimodal* embeddings; Google documents image, audio, and document input through the native `embedContent` API, not through the OpenAI-compatibility layer. Text is what this buys.

**What the PR should do:**

- **Read the credential from the environment, not from the YAML.** Follow `palinode/core/auth.py`'s `load_api_token()` (line 40) exactly: an env var, plus a `…_FILE` variant whose contents are the secret, stripped, logging an **error** when the file is set but unreadable — and, as that function does, without echoing the path, which can itself leak deployment topology. `PALINODE_EMBEDDING_API_KEY` / `PALINODE_EMBEDDING_API_KEY_FILE` matches the existing naming. **No key field in `palinode.config.yaml.example`** — a secret in a file people copy and commit is the failure this avoids; document the env var there instead.
- **Send `Authorization: Bearer <key>` only when the dialect is `openai` and a key is configured.** A key in the environment must never be attached to a request to a native Ollama endpoint — that would ship an operator's credential to a host that never asked for one. This wants its own negative test, not just a comment.
- **Make the endpoint path addressable** rather than a fixed `"/v1/embeddings"` suffix. The shape is yours to propose — an optional config field for the path, or treating the configured `url` as the full endpoint when it already names one — but say which you picked and why in the PR, and keep the existing bare-host and trailing-`/v1` forms working exactly as they do today.
- **Scope it to the EMBED role.** The chat and consolidation paths have the same missing credential, but their fallback chain needs a per-endpoint answer; that is a separate issue and this PR should not touch it.
- **Never log the key**, and make sure it cannot reach the structured request log the client already emits.
- **Tests:** the header is sent when the env var is set and the dialect is `openai`; **nothing is sent when the dialect is `ollama`, even with the env var set**; nothing is sent when no key is configured; the `…_FILE` variant is read and stripped; an unreadable `…_FILE` logs an error rather than crashing; the resolved URL is correct for a bare host, a `…/v1` host, and a provider root that is neither; and the key appears in no emitted log line. Use a stub transport — do not call a real provider in tests.
- **Docs:** extend the *Embeddings without Ollama* paragraph in `README.md` (line 380) with the hosted case and the env var, and add a commented hosted example to the `embeddings.primary` block in `palinode.config.yaml.example`.

**Out of scope, deliberately:** anything about vector *width*. `chunks_vec` is created as `FLOAT[{dimensions}]` from `embeddings.primary.dimensions`, so switching an existing store to a provider with different dimensions needs a reindex, and there is no mismatch guard today. Real problem, different issue — please don't fix it here.

Comment here to claim it. **Please wait to be assigned before opening a PR** — two contributors landed on one issue once because we had no such line.

Isha-Zaka:
## Summary

Adds bearer authentication and configurable endpoint paths for hosted
OpenAI-compatible embedding providers.

The existing OpenAI-compatible embedding path assumes `/v1/embeddings`
and had no way to provide credentials, which prevents providers with a
different API root or bearer-auth requirement from being used.

## Changes

- load embedding credentials from `PALINODE_EMBEDDING_API_KEY` or
  `PALINODE_EMBEDDING_API_KEY_FILE`
- send `Authorization: Bearer` only for the `openai` embedding dialect
- add optional `embeddings.primary.endpoint_path`
- preserve the existing bare-host and trailing `/v1` behavior when
  `endpoint_path` is not configured
- document hosted OpenAI-compatible providers and the environment variables
- add coverage for auth, `_FILE` loading, endpoint paths, Ollama isolation,
  and secret logging

I used an optional `endpoint_path` so existing configurations continue to
behave as before, while hosted providers can treat `url` as the provider root
and append their provider-specific embedding path.

Native Ollama never receives the embedding API key.

## Testing

- `pytest tests/test_embed_openai_dialect.py -v --tb=short`
- `pytest tests/test_api_bearer_auth.py -v --tb=short`
- `pytest -q`
- `bandit -r palinode/ -ll`
- `git diff --check`

Closes #203


Isha-Zaka:
@Paul-Kyle Thanks for catching this. I pushed a follow-up that redacts the credential from authenticated provider response bodies before they are carried into exceptions, including the 4xx and deterministic 5xx paths.

I also added a stub-transport regression test that drives a 401 response echoing the key through the embedder wrapper and verifies the key is absent from both the logs and the raised EmbeddingUnavailable.
Also documented the new embedding credential pair in SECURITY.md

For the _FILE behavior, the per-request resolution is intentional so file-backed credentials can be rotated without requiring a restart.

All checks are green now.

Paul-Kyle:
Merged — thank you for addressing the redaction gap and adding the regression test.

Two details worth naming. You redact **before** truncating to 300 characters; the other order looks identical in review and leaves a key fragment whenever the credential straddles the boundary. And your regression drives a credential-bearing 401 through the embedder wrapper rather than the client, then asserts `[REDACTED]` is present as well as the secret being absent — a test that only checks absence passes just as happily when the request never happened.

On per-request resolution of `PALINODE_EMBEDDING_API_KEY_FILE`: that is a good reason, so it stays, and it is now documented as intentional rather than looking like an oversight to whoever reads it next.

One limitation for the record, not a request: redaction matches the credential as a literal string, so a provider that echoes it re-encoded or partially masked would slip through. That is the right tradeoff for the failure we actually have, and worth knowing before anyone treats it as a guarantee.

PR #213, opened 2026-09-15
doctor: check that memory_dir is writable

### LINKED ISSUE #206: doctor: memory_dir_exists is critical but never checks the directory is writable
`memory_dir_exists` (`palinode/diagnostics/checks/memory_dir.py:16`) is registered `severity="critical"` and its module docstring says *"without it nothing works."* It asserts only that the path exists and is a directory:

```python
if memory_dir.exists() and memory_dir.is_dir():
    return CheckResult(..., passed=True, message=f"Memory directory exists: {memory_dir}")
```

**A memory directory that exists but is not writable passes this check**, and then every save fails. That is not hypothetical: a read-only bind mount, a volume whose ownership does not match the running user, or a directory whose mode was tightened by hand all produce exactly that state. `palinode doctor` is the "did it install correctly?" command, and on such a machine it reports the critical check green.

**The repo already solved this for a less important path.** `audit_log_writable` (`palinode/diagnostics/checks/audit_log.py:26`) checks the parent directory with `os.access(..., os.W_OK)` (line 94) and the resolved file (line 111), and reports a `warn` with remediation when it fails. The audit log gets a writability check; the directory that holds every memory does not.

**What the PR should do:**

- **Add a separate check, `memory_dir_writable`**, rather than widening `memory_dir_exists`. Two reasons: the existing check's name would stop describing what it does, and a separate result gives the user a distinct message and remediation — which is how `audit_log_writable` is structured.
- **Register it `@register(tags=("fast",))`**, matching its sibling (`memory_dir.py:15`); it is a stat call, not I/O.
- **Severity `critical`**, matching `memory_dir_exists`. An unwritable memory directory is not a warning — nothing can be saved.
- **Check `os.W_OK | os.X_OK`, not `W_OK` alone.** Creating a file inside a directory requires *search* permission on it as well as write, so a directory that is `-w-` is unusable in practice. (`audit_log_writable` checks `W_OK` alone on its parent — that is the same gap, not a weaker requirement it correctly applies. Fixing it is out of scope here; one check per issue.)
- **Mirror `audit_log_writable`'s mechanism otherwise**: `os.access(str(memory_dir), os.W_OK | os.X_OK)`. `os.access` is advisory and can disagree with an actual write — that trade is already accepted in this package, and a check that is right on every ordinary misconfiguration beats writing a probe file into the user's memory directory. **Do not add a write probe.**
- **Define the missing / not-a-directory case as an explicit pass, not a skip.** If `memory_dir` does not exist or is not a directory, `memory_dir_exists` already reports it, and a second failure for one cause is noise. Return `passed=True` with a message saying the check did not apply because the directory is absent — a *defined informational result*, not an undefined "skip".
- **Remediation text** should name the likely causes: ownership mismatch, a read-only mount, and a hand-tightened mode.
- **Tests:** a writable directory passes; a directory with write permission removed fails and carries remediation; a missing directory returns the informational pass rather than a second failure. Three constraints on the permission test, all of which have bitten people before:
  - **Restore the mode in a `finally`**, or `tmp_path` cleanup fails on some platforms.
  - **Skip when running as root** — `os.access` reports success for a privileged user regardless of mode, so the test would silently pass for the wrong reason. `os.geteuid() == 0` is the guard.
  - **Skip on Windows**, where POSIX mode bits do not govern this. `tests/` has `skipif` precedent for both.

**Out of scope:** `db_path`, the audit log, and anything about `--fix`. Registering a fix action requires an explicit safety-whitelist edit (`palinode/diagnostics/fixes.py`), and "make the user's memory directory writable" is not a change a tool should make on its own.

chiruu12:
Closes #206.

`memory_dir_exists` is registered `critical` and only asserts the path exists and is a
directory. A memory directory that exists but cannot be written to passes it, and then
every save fails. Confirmed on a directory with the write bit cleared:

```
memory_dir_exists: passed=True sev=critical
real write:        PermissionError
```

That is the state produced by a read-only bind mount, a volume whose ownership does not
match the running user, or a mode tightened by hand.

This adds a separate `memory_dir_writable` check rather than widening the existing one,
so the old name keeps describing what it does and the new failure gets its own message
and remediation. Same shape as `audit_log_writable`: `os.access`, no write probe.

Two details worth flagging.

It tests `os.W_OK | os.X_OK`, not `W_OK` alone. Creating a file inside a directory needs
search permission as well as write, so a directory that is `-w-` is unusable in practice.
`audit_log_writable` checks `W_OK` alone on its parent and has the same gap; left alone
here, one check per PR.

An absent or non-directory path is a defined pass, not a skip, with a message saying the
check did not apply. `memory_dir_exists` already reports that case and one cause should
not produce two failures.

Five tests, on real directories and real mode bits. The permission tests restore the mode
in a `finally`, skip on Windows, and skip under root, where `os.access` reports success
regardless of mode.

Docs: a row and a section in `DOCTOR.md`, and a CHANGELOG entry under Unreleased.

No `--fix` action. Registering one needs a `fixes.py` whitelist edit, and making a user's
memory directory writable is not a change the tool should make on its own.


Paul-Kyle:
Merged — thank you, and sorry it sat as long as it did.

Three things I want to name. Adding a check rather than widening `memory_dir_exists` keeps the old name honest and gives the new failure its own message, which is the difference between a doctor that reports and one that explains. Requiring `os.X_OK` alongside `W_OK` is correct and is the part a reviewer would have missed — a directory you can write to but not search is unusable, and you found that `audit_log_writable` has the same gap and left it alone anyway. One check per PR is the right instinct; open that one when you want it.

And skipping the permission tests under root is the detail people forget: `os.access` reports success there regardless of mode, so without the skip those two tests fail in any root container and the failure reads as a problem with the check rather than with the environment.

Issue #214, opened 2026-09-15
rel_path is backslash-separated on Windows: to_rel_path uses os.path.relpath and nothing normalises to POSIX form


Paul-Kyle:
`rel_path` is part of the API contract — `test_rel_path_api_parity.py` exists to hold it — and on Windows it is built from `os.path.relpath`, which returns backslash-separated paths.

**The chain.** `_enrich_with_rel_path()` (`palinode/api/search_helpers.py`) is what attaches `rel_path` to search, associative, dedup-suggest, orphan-repair and topic-coverage results. It calls `to_rel_path()` (`palinode/core/path_guard.py`), whose body is `os.path.relpath(file_path, base)`. That is correct POSIX behaviour and correct Windows behaviour — it is just not the same string, and **`as_posix()` appears nowhere in the package.**

**Seven `test_rel_path_api_parity` tests fail on native Windows** in @Kaap10's 3.12 inventory and @kevin-lozada-santos's 3.11 inventory, both posted on #169. Whether all seven trace to this expression is for the fix to establish — **that has not been verified, and it should not be assumed.**

**A separate, smaller instance in the same family.** `palinode/core/store.py` derives a path for associative results with `fp.replace(config.memory_dir + "/", "").lstrip("/")` — a hardcoded separator. On Windows that replace never matches, so the value stays absolute. It feeds the result's `category` field, **not** `rel_path`, so the consequence is a wrong category rather than a broken contract. Worth fixing in the same pass; worth not conflating with the above.

**The precedent is already in the tree.** `palinode/core/skip_dirs.py` does `rel_path.replace(os.sep, "/")`. Someone solved this once, in one place, and it was never generalised.

**What is wanted:** `rel_path` renders in POSIX form on every platform, and a test that pins it so the next refactor cannot quietly undo it. Where the normalisation belongs — inside `to_rel_path`, or at the enrichment boundary — is a design call worth stating in the PR.

**Not in scope:** the rest of the Windows suite failures. Those are being carved separately.

**Context:** found by @Kaap10 and @kevin-lozada-santos during the native Windows runs on #169.

**Reserved for @Kaap10**, who claimed this slice on #169. GitHub currently rejects the direct assignment on this new issue; it is not open for another claim.

Kaap10:
I've implemented the fix and opened a pull request to resolve this issue: #228

### Summary of Changes:
[...]

PR #215, opened 2026-09-15
fix(worktree): use non-signalling Win32 API for _pid_alive on Windows (#212)

### LINKED ISSUE #212: worktree-reconcile: the _pid_alive probe can signal or terminate a process on Windows instead of testing whether it exists
`_pid_alive` in `palinode/cli/worktree.py` decides whether a worktree's owning process is still running, and `palinode worktree-reconcile` removes worktrees on a `False`:

```python
try:
    os.kill(pid, 0)
    return True
except ProcessLookupError:
    return False
```

On POSIX, signal `0` is the documented way to ask "does this process exist" without delivering anything. **On Windows it is not a query at all.** CPython's `os.kill` treats `0` as `CTRL_C_EVENT` and attempts `GenerateConsoleCtrlEvent`; **if that call fails it falls through to `OpenProcess` + `TerminateProcess`.** So the probe either delivers a console control event to the target or tries to kill it — and which of those happens depends on console attachment, not on anything we control.

**The function's own docstring states the invariant this breaks:** a `PermissionError` means the process exists but belongs to another user, and any other OS error means we cannot tell — "both are treated as alive so we never remove a worktree whose owner might still be running." On Windows it is not asking that question.

**Observed on two hosts and two Python versions.** @kevin-lozada-santos on Windows 11 / Python 3.11.15 and @Kaap10 on Windows 11 / Python 3.12.7 both report the suite stalling at `tests/test_worktree_reconcile.py`. **Eight tests are common to both inventories** as having no completed native result:

```
test_pid_alive_self_true_and_dead_false
test_dead_clean_upstream_is_removed
test_alive_lock_is_skipped
test_dirty_tree_is_skipped
test_no_upstream_is_skipped
test_apply_removes_only_dead_clean_upstream
test_cli_dry_run_default_removes_nothing
test_cli_execute_removes
```

These are tests the runner never finished after the stall, not eight independently diagnosed failures. Both inventories are posted in full on #169.

**What is wanted:** a **non-signalling** existence query on Windows — `OpenProcess` with `PROCESS_QUERY_LIMITED_INFORMATION`, reading the exit code, and closing the handle — while POSIX keeps `os.kill(pid, 0)` unchanged. **Preserve the fail-safe direction:** anything undeterminable must still count as alive, because the caller deletes worktrees on a `False`.

**Tests:** three outcomes, and none of them may signal the target — **live**, **dead**, and **unknown/undeterminable**, with the last resolving to *alive*. That third one is the property the docstring promises and the one whose failure deletes someone's work. Make sure any handle the probe opens is closed on every path, including the error paths.

**Not in scope:** the other Windows suite failures. Those are #169's surface and are being carved separately.

**Credit:** reported by @kevin-lozada-santos, independently observed on 3.12 by @Kaap10.

**Reserved:** @kevin-lozada-santos and @Kaap10 have first refusal on this one, since it came out of their runs. If neither has claimed it by **21 September 2026 at 14:00 UTC**, it is open to anyone — comment here and **please wait to be assigned before opening a PR.**

Kaap10:
## Summary
Fixes #212.

On Windows, `os.kill(pid, 0)` maps to `CTRL_C_EVENT`, which signals the console process group instead of probing process existence without side effects.

This PR replaces the Windows probe with a non-signalling Win32 API query using `OpenProcess` with `PROCESS_QUERY_LIMITED_INFORMATION | SYNCHRONIZE` (0x00101000) and `WaitForSingleObject(handle, 0)`:
- `WAIT_TIMEOUT` (0x102) -> Process is actively running (`True`).
- `WAIT_OBJECT_0` (0x0) -> Process has exited / signaled (`False`).
- Fully immune to the Windows `259 (STILL_ACTIVE)` exit-code trap.
- Preserves the fail-safe invariant: `ERROR_ACCESS_DENIED (5)` and unexpected OS errors resolve to `True` (alive); `ERROR_INVALID_PARAMETER (87)` resolves to `False`.
- Hardens input boundaries (rejects booleans, non-ints, and PIDs `> 0xFFFFFFFF`).
- Normalizes path casing in `_under_claude_worktrees` on Windows using `os.path.normcase`.
- Preserves POSIX `os.kill(pid, 0)` completely unchanged.

## Test Plan
- Added comprehensive unit tests in `tests/test_worktree_reconcile.py


Paul-Kyle:
This is strong work and I want to name the best decision in it before the one thing I am holding on.

You didn't take the issue's suggestion to read the exit code, and you were right not to: `GetExitCodeProcess` returns 259 for a running process, so a process that genuinely exits with code 259 would read as alive forever. `test_pid_alive_exit_code_259_dead` spawns exactly that case and pins it. That turns a deviation from the spec into a documented necessity, and it is what will stop someone simplifying it back in a year.

**What I am holding on: the error capture.** `ctypes.windll.kernel32.GetLastError()` is a separate foreign call made after `OpenProcess` returns, so there is no guarantee that what it reports is the error `OpenProcess` set. I originally wrote this off as safe on the grounds that a *lost* 87 falls through to alive — but that only covers one direction. The other direction is the one that matters: if the value observed is 87 when the real failure was access-denied, a process that is alive reads as dead, and `reconcile` is then free to remove its worktree. That is the invariant this whole issue exists to protect, and I should not have waved it through on a guarantee I hadn't established.

The fix is small: `ctypes.WinDLL("kernel32", use_last_error=True)` and `ctypes.get_last_error()`, which captures the value immediately after the foreign call into a per-thread slot where nothing can clobber it. The existing error-code tests will need to patch the new accessor.

**To be clear about what I have and haven't done:** I have not reproduced this on Windows, and I'm not claiming an intervening call definitely overwrites the value. The problem is narrower — the current form gives no guaranteed immediate capture, the code's correctness depends on having it, and there is a documented idiom that provides it.

**One smaller thing for the same push.** `test_pid_alive_win32_wait_codes` has its whole body inside `if sys.platform == "win32":` with no `else`, so on our Ubuntu and macOS runners it executes nothing and still reports green. `pytest.mark.skipif` would make that visible as *skipped* rather than passing. That matters here more than usual, because our CI has no Windows lane at all — the 13/13 on this PR says nothing about the branch you actually wrote, and your own 15/15 native run on #212 is the real evidence for that path. I'd rather the test output stopped implying otherwise.

`test_pid_alive_failsafe_on_access_denied_or_error` is **not** the same case — its `else` branch genuinely exercises the POSIX `PermissionError` path, so please don't mark that one Windows-only. Splitting it into a Windows test and a POSIX test would be tidiest, but keeping the `else` exactly as it is works too.

Everything else stands: fail-safe on every wait branch including `WAIT_FAILED`, `CloseHandle` asserted after each outcome, live and dead exercised against a real child process, and an input guard that quietly fixes POSIX too — `os.kill(-1, 0)` was a permission probe against every process on the machine. The `normcase` change rode along outside scope and I'd rather have it than not.

Push onto the same branch and I'll review and merge once checks are green.

Kaap10:
@Paul-Kyle Thanks for catching that - using `use_last_error=True` with `ctypes.get_last_error()` makes complete sense to guarantee we never lose the error code to an intervening call.

Also updated the test suite with `pytest.mark.skipif` and split the access-denied tests for Windows and POSIX so platform skips are transparent in CI.

Pushed commit `00a0b16` - ready for re-review!

Paul-Kyle:
The code change is exactly right, and I want to be clear that what I am asking for now is my mistake, not yours.

Deleting the `GetLastError` argtypes and restype rather than leaving them declared-but-unused is the detail I'd have let slide. Splitting the fail-safe test into Windows and POSIX rather than marking the whole thing Windows-only kept the POSIX `PermissionError` coverage alive — the lazier reading of my note would have deleted it. And the win32 test now patches `ctypes.get_last_error`, which is the accessor the code actually calls; patching the old symbol would have left a test that passed while testing nothing.

**The one thing I need before merging is a native run of the new head.** Your 15/15 on #212 was posted just before `af4e5d92` was committed, so it reflects an earlier state of the branch. `00a0b165` came a day later and changed the Win32 branch specifically — the `WinDLL`/`get_last_error` conversion and the test that exercises it. Our CI has no Windows runner, so those tests report as *skipped*, and there's no documented native run of this head on the thread. I cited your earlier run as evidence for it in my own notes, which was wrong of me — it's evidence from before this change.

Could you run `tests/test_worktree_reconcile.py` on native Windows at `00a0b165` and paste the result? That file alone is enough; I'm not asking for the full suite — and if you already ran it, the paste is all I need. The reason I'd rather have it than wave it through: the fail-safe tests patch `ctypes.get_last_error` directly, so they never exercise the capture itself. If the conversion were wired wrong, the value read wouldn't reliably reflect `OpenProcess`, and the unit tests couldn't tell. This is the one probe that decides whether a worktree gets deleted, so I'd like it seen running once.

Everything else stands and I'll merge on the result.

Kaap10:
@Paul-Kyle Here is the native Windows run receipt on commit `00a0b165`:

**Environment:** Windows 11 (build 26200), Python 3.12.7, pytest 9.1.1

```text
============================= test session starts =============================
platform win32 -- Python 3.12.7, pytest-9.1.1, pluggy-1.6.0 -- C:\Users\vardh\palinode\.venv\Scripts\python.exe
cachedir: .pytest_cache
metadata: {'Python': '3.12.7', 'Platform': 'Windows-11-10.0.26200-SP0', 'Packages': {'pytest': '9.1.1', 'pluggy': '1.6.0'}, 'Plugins': {'anyio': '4.15.1', 'asyncio': '1.4.0', 'json-report': '1.5.0', 'metadata': '3.1.1', 'timeout': '2.4.0'}}
rootdir: C:\Users\vardh\palinode
configfile: pyproject.toml
plugins: anyio-4.15.1, asyncio-1.4.0, json-report-1.5.0, metadata-3.1.1, timeout-2.4.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
timeout: 60.0s
timeout method: thread
timeout func_only: False
collecting ... collected 16 items

tests/test_worktree_reconcile.py::test_pid_alive_self_true_and_dead_false PASSED [  6%]
tests/test_worktree_reconcile.py::test_pid_alive_child_process_lifecycle PASSED [ 12%]
tests/test_worktree_reconcile.py::test_pid_alive_failsafe_on_access_denied_or_error_win32 PASSED [ 18%]
tests/test_worktree_reconcile.py::test_pid_alive_failsafe_on_access_denied_posix SKIPPED [ 25%]
tests/test_worktree_reconcile.py::test_pid_alive_invalid_types_and_bounds PASSED [ 31%]
tests/test_worktree_reconcile.py::test_pid_alive_win32_wait_codes PASSED [ 37%]
tests/test_worktree_reconcile.py::test_pid_alive_exit_code_259_dead PASSED [ 43%]
tests/test_worktree_reconcile.py::test_under_claude_worktrees_case_and_boundary PASSED [ 50%]
tests/test_worktree_reconcile.py::test_parse_porcelain_extracts_locked_and_branch PASSED [ 56%]
tests/test_worktree_reconcile.py::test_dead_clean_upstream_is_removed PASSED [ 62%]
tests/test_worktree_reconcile.py::test_alive_lock_is_skipped PASSED      [ 68%]
tests/test_worktree_reconcile.py::test_dirty_tree_is_skipped PASSED      [ 75%]
tests/test_worktree_reconcile.py::test_no_upstream_is_skipped PASSED     [ 81%]
tests/test_worktree_reconcile.py::test_apply_removes_only_dead_clean_upstream PASSED [ 87%]
tests/test_worktree_reconcile.py::test_cli_dry_run_default_removes_nothing PASSED [ 93%]
tests/test_worktree_reconcile.py::test_cli_execute_removes PASSED        [100%]

======================= 15 passed, 1 skipped in 18.06s ========================
```

The unmocked dead-PID check in `test_pid_alive_self_true_and_dead_false` exercises the live `WinDLL("kernel32", use_last_error=True)` call and captures error code 87 (`ERROR_INVALID_PARAMETER`) end-to-end. Ready for merge!

Paul-Kyle:
Merged — thank you, and thank you for the receipt.

The detail I checked before merging was your note about the dead-PID test, and it holds: `test_pid_alive_self_true_and_dead_false` calls `_pid_alive(999_999)` unmocked, and the code only returns `False` there when `ctypes.get_last_error()` reports 87. Anything else falls through to alive and the assertion fails. So your native pass shows the `use_last_error` capture working end to end, not just the mocked branches. That's exactly the evidence that was missing, and pasting the full pytest header with platform and interpreter made it checkable rather than asserted.

For the record: this went through three rounds, and two of the holds came from me — first the error-capture reasoning, then citing your earlier run as evidence for a later commit. You answered both quickly and without argument. The `WaitForSingleObject` choice and `test_pid_alive_exit_code_259_dead` are still the best part of it.

#214 is reserved for you whenever you want it.

Issue #169, opened 2026-08-30
Native Windows: run the suite and triage what remains


Paul-Kyle:
Successor to https://github.com/phasespace-labs/palinode/issues/93, which is being split and closed as this is filed. This carries what is genuinely unfinished: **nobody has run the suite on native Windows.**

**What #93 settled, so this issue does not re-litigate it:**

- Test helpers that read Palinode-written files pass `encoding="utf-8"` — done across four areas by @chiruu12 (#149, #152, #153, #167), with all swept files joining a ratcheting guard so the area cannot regress.
- The `os.fchmod` overwrite path has a Windows **fallback implemented** — `getattr(os, "fchmod", None)` with a path-based `os.chmod` when it is absent (`palinode/core/git_tools.py:191-196`, from #94). Its *behaviour on native 3.11/3.12 has never been observed*, which is item 2 below.
- `write_memory_file`'s directory fsync is platform-guarded.

**What is left:**

1. **Run the suite on native Windows** — not Docker — and post what fails. That is the whole ask. A failure list is a complete contribution here; fixing what it finds can be separate issues.
2. Confirm the `os.fchmod` fallback actually behaves on 3.11 and 3.12, rather than on the assumption that the `getattr` guard is sufficient.

**This needs hardware we do not have**, which is why it is unassigned and why it may sit. Windows is not in CI today — Docker is the supported Windows path — so nothing here affects shipped behaviour. It does mean the suite cannot be run natively on Windows, which closes the door on Windows contributors.

A `windows-latest` CI lane would be the durable fix and is deliberately **out of scope** — this issue is about making the suite pass there first. Say so on the thread if you want to take that instead.

Kaap10:
Hi @Paul-Kyle, I am on native Windows (Windows 11, Python 3.12). 

I'd like to take this on - I will run the full test suite natively on Windows, triage the failures (such as path separator assertions and encoding issues), and submit fixes so the suite passes cleanly on native Windows.

Please assign this to me!

kevin-lozada-santos:
I can cover the **native Windows / Python 3.11** side to complement @Kaap10's announced 3.12 run. I'll post the observed suite failures and exercise the real `os.fchmod` fallback on this host. This contribution will be test results and triage; I won't compete with the ongoing implementation work.

kevin-lozada-santos:
[...]

PR #217, opened 2026-09-16
fix(core): preserve read-only write failures during Windows cleanup

### LINKED ISSUE #216: write_memory_file: overwriting a read-only target on Windows strands the temp file and replaces the original error
When `write_memory_file` overwrites an existing target that is read-only, `os.replace` raises `PermissionError`. The original file survives intact — that part is correct. But on native Windows **the temporary file is left behind**, and the exception the caller sees is not the one that stopped the write.

**Observed on two hosts.** Both native Windows runs on #169 recorded it: @kevin-lozada-santos on Python 3.11.15 and @Kaap10 on 3.12.7 — *"one `.tmp` file remains pending cleanup … original contents survive."*

**Probable mechanism** — read from the code, to be confirmed by your regression rather than taken on trust. Before writing, the function copies the **target's** mode onto the temp file (`mode = os.stat(file_path).st_mode & 0o777`, then `fchmod` or, on Windows, `os.chmod(tmp_path, mode)`). A read-only target therefore produces a read-only temp file. When `os.replace` then fails, the cleanup block calls `os.unlink(tmp_path)` and catches only `FileNotFoundError` — but on Windows, unlinking a file with the read-only attribute raises `PermissionError`. That escapes the `except`, **replaces the original exception**, and the temp file stays on disk. On POSIX the same code is fine, because unlink is governed by the directory's permissions rather than the file's.

**What is wanted**, in the scope you proposed on #169:

- When replacing a read-only destination fails, **preserve the destination's contents and attributes** — already true; keep it true.
- **Remove only the temporary file created by that write attempt.** On Windows that likely means clearing the read-only attribute on the temp file before unlinking it.
- **Keep the original failure meaningful to the caller.** The exception that propagates should be the one raised by `os.replace`. If cleanup itself fails, that is worth a log line — not a replacement exception.

**Please keep the fix inside the failure path.** Deferring the mode copy until after `os.replace` succeeds would trade this bug for a different one: a write that has already replaced the target and then fails to restore its permissions, which is harder to reason about than a stranded temp file. The cleanup block is the right place.

**Tests:** a native Windows regression that **fails on the current implementation** — a read-only target, an attempted overwrite, then assert that no `.tmp` remains in the directory and that **the original exception survives**. Assert on identity rather than on message text: `os.replace` sets both `filename` (the temp source) and `filename2` (the destination), so a correct error legitimately names both, and an assertion that it names one "rather than" the other would be wrong. What matters is that the error the caller catches is the one `os.replace` raised, not a second one from cleanup. Please also confirm the normal writable-overwrite path still leaves no temp file and still writes byte-for-byte — both #169 receipts verified that by hand.

**Scope:** `palinode/core/git_tools.py`'s `write_memory_file` and its tests. Nothing else in that module.

**Reserved for @kevin-lozada-santos**, who found it and proposed the scope above on #169. It is not open for another claim. Comment here when you are ready and I will sort the assignment out.

kevin-lozada-santos:
Fixes https://github.com/phasespace-labs/palinode/issues/216.

A failed overwrite of a read-only destination on Windows can also fail to unlink the temporary file whose mode was copied from that destination. Cleanup then strands the temporary file and masks the original replacement exception.

This change retries deletion after making only that temporary file writable on Windows. Cleanup errors are logged and the original write exception is re-raised. The successful-write path and pre-rename mode-copy order stay unchanged; the destination is never made writable.

The native Windows regression uses the real `os.replace`, captures its exception, and asserts that the exact same object is raised to the caller. It checks unchanged destination content and mode, preservation of an unrelated read-only temporary-file sentinel, and removal of only the write's temporary file. Additional tests cover exact UTF-8 bytes and mode for writable overwrites, plus cleanup failures that must be logged without masking the original error.

Validation on native Windows / CPython 3.11.15:

- New regression file on unchanged upstream `d4379d87d7f13f041ef3ddeca18cab2751ee471b`: **3 failed, 1 passed**.
- Same file on this commit: **4 passed**.
- Focused git-tools, UTF-8, executor, fact-ID, and issue-reference checks: **118 passed, 3 xfailed, 1 failed**. The failure, `test_atomic_write_directory_fsync_failure_propagates_and_cleans_temp`, also fails on unchanged upstream: it expects a directory fsync error while Windows skips directory fsync. No broader-suite green claim is intended.
- `ruff check palinode/ tests/ scripts/`: passed.
- `bandit -r palinode/ -ll`: passed.

Full suite not rerun; unrelated native process-liveness tests remain outside this case. Changelog entry is under the existing Unreleased / Fixed heading.

Commands for reproduction:

```text
python -m pytest tests/test_git_tools_write_cleanup.py -q --tb=short
python -m pytest tests/test_git_tools.py tests/test_git_tools_write_cleanup.py tests/test_utf8_encoding_guard.py tests/test_executor.py tests/test_executor_ops_coverage.py tests/test_executor_replace_guard.py tests/test_fact_ids.py tests/test_no_issue_refs_user_surface.py -q --tb=short
```


Paul-Kyle:
Merged — thank you. This is the scope you proposed on #169, built exactly as described.

The test is the part I want to name. Asserting `caught.value is replace_errors[0]` — the same exception object, not a message match — is the assertion that actually pins "the original failure survives cleanup", and it sidesteps the trap that `os.replace` legitimately names both the temp source and the destination. Checking `filename2` as an attribute rather than parsing the string is the same instinct.

The read-only sentinel is better than the obvious test. "No `.tmp` remains" would pass for a fix that swept the directory; proving an unrelated read-only temp file survives with its mode intact is what shows the blast radius is one file.

One thing worth having on the record: on POSIX the re-raise inside the inner handler now lands in the outer `except OSError` and gets logged, so an unlink failure there stops masking the original error too. If that was deliberate, good — it is the same bug in the quieter platform.

And thank you for how you reported the validation. You named the platform and interpreter, scoped the claim to the checks you actually ran, and flagged that `test_atomic_write_directory_fsync_failure_propagates_and_cleans_temp` fails on unchanged upstream too rather than letting it read as something you broke — then said plainly that no broader-suite green claim was intended. Our CI has no Windows lane, so on Windows changes that report *is* the evidence, and a report that marks its own edges is worth more than a bigger number.

Issue #189, opened 2026-09-04
Prevent prompt activation from leaving two prompts active after a failed deactivation


Paul-Kyle:
`palinode/api/routers/session.py`, `activate_prompt_api()`. The loop that deactivates other prompts of the same task exists to enforce one active prompt per task:

```python
# Deactivate all prompts of the same task
for filepath in glob.glob(os.path.join(prompts_dir, "*.md")):
    try:
        ...
        if info["task"] == task and resolved != target_path:
            _set_active(resolved, False)
    except Exception:
        pass

# Activate target
_set_active(target_path, True)
```

**`_set_active(target_path, True)` runs unconditionally, after the loop.** If deactivating one sibling raises — a malformed frontmatter block, a write that fails — that prompt stays `active: true`, the target is activated anyway, and the task ends up with **two active prompts**. The endpoint returns success.

The swallowed exception is no longer silent: #179/#181 added a warning to the sibling loop in `list_prompts_api`, and this loop's `except Exception: pass` is what remains. But logging was deliberately scoped out of that change, because **making the failure audible does not stop the bad state** — the control flow does.

**There is an open design question, and it should be settled in this thread before anyone writes code:**

- **Fail closed** — if any deactivation fails, do not activate the target; return an error describing which file blocked it. Safest, but a single unreadable file makes activation impossible until it is fixed by hand.
- **Roll back** — re-activate whatever was already deactivated, then report failure. Restores the invariant, but the rollback writes can themselves fail, and a partial rollback is a worse state than either endpoint.

**Please comment with a preference before implementing.** Whichever wins wants a test over partial-failure orderings — failure on the first sibling, on a middle one, and on the last — asserting that no two prompts for one task are ever left `active: true`.

**Scope:** `activate_prompt_api` and its tests. `list_prompts_api` is already done; leave it alone.

chiruu12:
Preference: fail closed.

Rollback has to write to the same files that just failed, so the failure mode it is meant to cover is the one most likely to break it, and a partial rollback leaves an invariant nobody can reason about. Fail closed has one outcome: the task keeps whatever was active before, and the error names the file that blocked it. Activation is user-initiated and retryable, so making it refuse until a malformed prompt is fixed seems like the right trade against silently ending up with two active prompts.

[...]

PR #218, opened 2026-09-16
test: widen the issue-ref guard to all of palinode/

### LINKED ISSUE #211: Eight private-tracker issue references ship outside palinode/diagnostics/, and the #207 guard does not reach them
#207 removed ten bare private-tracker references from `palinode/diagnostics/` and added a guard scoped to that package. **Eight more live outside it**, in six string constants across four files. Each was checked against this repository's tracker; all eight are wrong here. They all resolve on our private development tracker, which is why they survived review — they look correct to whoever wrote them.

**Locators are by symbol and quoted text on purpose.** Most of these sit inside long multi-line string constants; symbol-and-quote locators remain useful as lines move.

**Three outcomes, because they are not the same change.**

**1. `palinode/cli/mcp_config.py`, in the `mcp_config` command — repoint, do not delete.** This prints to the terminal in a red `WARNING` when your MCP configs diverge:

> Editing the wrong one is the silent-failure pattern documented in #189.

On this repository #189 is an unrelated issue about prompt activation, so the sentence sends a reader somewhere wrong rather than nowhere. **Point it at `docs/MCP-CONFIG-HOMES.md` instead** — that document ships, it is the canonical description of this failure, and `palinode mcp-config --diagnose` already surfaces it.

**2. `palinode/cli/init.py` ×5 — delete the references.** These are inside string constants written into *your* repository by `palinode init`:

- `HOOK_SCRIPT`: "…though it were what the session was about (#682). Strip it here, at the…"
- `WRAP_COMMAND_BODY`: "…explicit yes in this session (#618)." · "…so the note actually reaches the remote (#378). Without `push: true`…" · "…interrupted (#353). If the push succeeds, continue…"
- `WRAP_HEAVY_COMMAND_BODY`: "…would sit unpushed despite a \"heavy\" wrap (#378)."

A user should not end up owning files that cite our private issue numbers. Each sentence reads correctly without the number; **delete rather than replace, and do not substitute a URL** — a link into a private tracker is just as unusable.

**3. `palinode/cli/_api.py` and `palinode/mcp.py` — delete the references.** Both are the message of a module-level cross-surface drift assertion, identical but for the filename:

> …without PALINODE_SESSION_END_TIMEOUT override — update cli/_api.py or defaults.py to stay in sync (#377)

The message already says what to do; the trailing `(#377)` adds nothing a reader can act on.

**Then extend the guard.** `tests/test_no_issue_refs_user_surface.py` already has `test_no_issue_refs_in_diagnostics_strings`, which walks `palinode/diagnostics/` and fails on bare refs in string constants. **Widen the scanned root to `palinode/`** so the rest of the package is covered. Its pattern is `#\d{2,4}\b`, which is already narrow enough to ignore the single-digit finding numbers in `api/server.py` and `api/enrichment.py` ("Tier B finding #3", "Tier B #5") — those are not tracker references and need no exemption. Keep `tests/` out of scope; it carries deliberate fixtures, including this file's own `test_bare_number_is_rejected`.

**Verify it bites:** add a bare `#1234` to one of the fixed strings, watch the guard fail, then remove it. A guard nobody has watched fail is a guard nobody knows works.

**Scope:** four source files, plus `tests/test_no_issue_refs_user_surface.py`. **No behaviour change** — reference strings, one doc pointer, and the guard's scope.

**Context, not scope:** #196/#204 fixed this class in the parity registry, #207 in the diagnostics package. This is the last of it in `palinode/`.

chiruu12:
Closes #211.

Widens `test_no_issue_refs_in_diagnostics_strings` from `palinode/diagnostics/` to `palinode/`, and clears the eight bare refs that widening exposes.

The guard was scoped to diagnostics because remediation text is printed straight at the user. That is the loudest case but not the only one. The three kinds of string it was missing:

- two import-time assertion messages, `cli/_api.py:22` and `mcp.py:385`, both `(#377)`
- a CLI divergence warning, `mcp_config.py:584`, `#189`
- five in the `/wrap` command body in `cli/init.py`, which `palinode init` renders into the user's own repo

Refs are deleted rather than swapped for a URL. Each one sat at the end of a sentence that already says what it needs to, so the number was the only part a public reader could not follow.

`mcp_config.py:584` is the exception. "the silent-failure pattern documented in #189" loses its referent if the number just goes, so it now points at `docs/MCP-CONFIG-HOMES.md`, which is where that pattern is written up.

`tests/` stays out of scope. Test strings are fixtures and assertion prose for contributors, and the pinned cases at the bottom of the guard file quote bare tags deliberately.

Checked the guard actually bites: dropped a bare `#1234` into the `mcp_config.py` warning string and the widened test failed on it, naming the file and line. The old diagnostics-only scope never walked that file at all. Removed it again.

Full suite: 3050 passed, 1 failed. The failure is `test_op_parse.py::test_parse_malformed_json_recovered_and_filtered`, which fails the same way on `main` with this branch stashed.


Paul-Kyle:
Merged — thank you, and thank you for the docstring.

Renaming the guard was the call I asked for, but writing down *why `tests/` stays out* is the part that will still be doing work in a year. The next person to widen this will read that paragraph instead of discovering the fixture at the bottom of the file the hard way.

Two other things. Repointing `mcp_config.py` at `docs/MCP-CONFIG-HOMES.md` rather than deleting the reference is right — that sentence is the one that genuinely loses its meaning without a target, and the doc ships. And you cleared the reference in `examples/hooks/palinode-session-end.sh`, which neither the widened string guard nor the comment guard can see: it is outside `palinode/` and it is not a `.py`. Fixing something the guard you just wrote cannot enforce is the harder habit.

PR #225, opened 2026-09-19
fix: preserve colliding vault wikilink targets

### LINKED ISSUE #221: import --from-vault: wikilinks are rewritten to the wrong note when two source files share a slug
`palinode import --from-vault` disambiguates destination paths when two notes slugify to the same name, but the map it uses to rewrite `[[wikilinks]]` is not disambiguated alongside them. Links end up pointing at the wrong note, silently.

**Reproduction.** A vault with three notes:

```
Projects/My Note.md     →  my-note.md
Projects/my-note.md     →  my-note-2.md
Projects/linker.md      →  body: See [[my-note]] and [[My Note]].
```

`[[my-note]]` and `[[My Note]]` name two different notes. After `palinode import --from-vault`, `linker.md` reads:

```
See [[my-note-2]] and [[my-note-2]].
```

Both point at the same file, and `[[My Note]]` points at the wrong one — its target kept the clean `my-note` stem.

**Where it happens.** In `plan_import` (`palinode/import_/vault.py`), destinations are computed first and `_make_dest_path` disambiguates them properly against a `used_paths` set. The slug map is then built by iterating the same list and assigning `slug_map[slugify(stem)] = dest_path.stem`. Two sources with the same slugified stem write the same key, so **the last one wins** and the earlier note's entry is lost. `_translate_wikilinks` then finds a hit for both spellings and rewrites both to the surviving value.

**No warning fires.** `orphan_warnings` only collects links with *no* match. Both lookups here succeed, so the import summary reports nothing, and `palinode orphan-repair` — which the module docstring recommends for post-import cleanup — cannot detect it either, because the link is not orphaned. It resolves. To the wrong note.

**What is wanted.** Wikilink rewriting that agrees with the destination disambiguation, and an honest report when it cannot. Two decisions are yours to make and worth stating in the PR:

1. **What should an ambiguous link resolve to?** An exact-stem match beating a slug match is one defensible answer; treating genuine ambiguity as unresolvable and leaving the link untouched is another. Whichever you choose, **ambiguity should be reported** rather than resolved by iteration order.
2. **Where the map should be keyed.** Keying by something unique per source — rather than by slugified stem — makes the collapse structurally impossible instead of handled.

**Tests.** `tests/test_import_vault.py::test_slug_disambiguation` already builds this collision and asserts only that the destination paths differ, which is why the defect survived. **Extend it, or add a sibling, that asserts on the translated body** — that two links naming two different sources resolve to two different destinations. Please also confirm the single-note case still rewrites as it does today (`test_wikilink_translation_within_import_set`) and that genuine orphans are still reported (`test_wikilink_orphan_left_as_is`).

**Scope:** `palinode/import_/vault.py` and its tests. No CLI change needed.

**How to see it:** the three-file vault above, then read the imported `linker.md`.

NesrineGharbi77:
## Summary

Fixes https://github.com/phasespace-labs/palinode/issues/221.

Vault imports no longer collapse wikilink targets when two source files slugify to the same stem.

## Design choices

1. **Exact source-stem matches beat slug matches.** For `My Note.md` and `my-note.md`, `[[My Note]]` resolves to the former and `[[my-note]]` resolves to the latter.

2. **The canonical destination map is keyed by unique source-relative path.** Slug and exact-stem indexes contain source keys rather than destination strings, so collisions cannot overwrite one another.

If a link has no unique exact match and its slug matches multiple sources, the link is left untouched and an ambiguity warning is added. Slug collisions are also reported even when all links happen to resolve exactly.

Collision/ambiguity reports use a separate `collision_warnings` return channel from true `orphan_warnings`. The CLI renders them under their own heading, so orphan counts and the `orphan-repair` guidance remain accurate. The global collision report is intentionally emitted even when no current wikilink is ambiguous: same-slug source namespaces should not become silent again.

## Tests

- Extended `test_slug_disambiguation` to assert translated wikilink bodies for colliding source stems.
- Added a regression test proving a genuinely ambiguous slug-only spelling is left unchanged and reported.
- Existing single-target translation and orphan behavior remain covered by their existing tests.

## Changelog

Added an `Unreleased / Fixed` entry for the behavior change.


Paul-Kyle:
Reviewed against `main`, and I reproduced the behaviour below on your branch rather than reading it off the diff.

**The fix is right, and you took the stronger of the two options.** On the three-note reproduction from the issue, `[[my-note]]` now lands on `my-note-2` and `[[My Note]]` on `my-note` — two spellings, two sources, two destinations. Keying the canonical map by source-relative path means the collapse cannot happen rather than being handled after the fact, which is what I was hoping someone would pick. Both decisions stated in the PR body, which is what the issue asked for.

**One change, and it is about where the report goes rather than whether to report.** The collision message is appended to `orphan_warnings`, and that list has exactly one consumer — `palinode/cli/import_vault.py` prints it as:

```
Orphaned wikilinks (N) — left as-is:
  warning: ...
  Tip: run `palinode orphan-repair` after import to fix unresolved links.
```

I ran your branch against a vault containing `My Note.md` and `my-note.md` **and no wikilinks at all**. It reports one warning — so the user is told a wikilink was orphaned and left as-is, and sent to `orphan-repair`, in an import with no wikilinks in it. `N` also stops being a count of orphans.

Reporting the collision is correct and I asked for it. Could you give it its own channel? A second returned list with its own heading in `import_vault.py` is the obvious shape, though if you would rather keep one list and carry a kind on each entry, that reads fine too — your call, just say which in the PR.

**One question, no action needed yet:** the global collision report fires once per colliding slug even when every link resolves exactly. I think that is right — a silent collision is what caused this bug — but it does mean a clean import of a vault with two same-slug filenames is never silent. If you meant that, say so and I will leave it alone.

Also: this is still marked draft. If it is ready, flip it — I read it as finished.

NesrineGharbi77:
Thanks — that makes sense. I’m going to split collision reporting out of `orphan_warnings` into its own returned list and give it a separate heading in `import_vault.py`, so orphan counts and the `orphan-repair` tip stay truthful.

And yes, the global collision report is intentional: I want same-slug source collisions to be visible even when every current wikilink happens to resolve exactly, because otherwise the import can silently retain a structurally ambiguous namespace again.

I’ll update the tests to cover both reporting channels, rerun the full checks, then mark the PR ready for review.

NesrineGharbi77:
Updated as discussed: collision/ambiguity reporting now has its own `collision_warnings` channel and CLI heading, while true orphan warnings keep their own count and `orphan-repair` guidance. I also added coverage for a same-slug vault with no wikilinks, so this exact reporting case is pinned. The global collision report remains intentional even when current links resolve exactly. All CI, wheel, plugin, lint, security, and platform unit-test checks are green, and I’ve marked the PR ready for review.

Paul-Kyle:
Merged — and thank you for the turnaround on that.

I re-ran the case I reported against your updated branch: a vault with two same-slug filenames and no wikilinks at all now returns zero orphan warnings, so the orphan count and the `orphan-repair` tip stay truthful. Pinning that exact case in the tests is what makes it stay fixed.

For the record of what landed: exact source-stem matches beat slug matches, the canonical map is keyed by source-relative path so the collapse cannot recur, and collisions report on their own channel — including when every current link happens to resolve, which you argued for and which I agree with.

PR #226, opened 2026-09-19
test: add issue-ref guard for examples directory

### LINKED ISSUE #224: test: the issue-ref guard does not cover examples/, so a hand-fixed ref there can silently regress
`examples/` ships publicly, and the guard's whole premise applies to it: palinode is developed in a private repo whose issue numbers are meaningful, so a bare `#NNNN` that ships resolves, for a public reader, to a different issue or to none at all. That is what `tests/test_no_issue_refs_user_surface.py` exists to prevent — and **`examples/` is outside every scan it runs.**

This is not hypothetical: a bare ref in `examples/hooks/palinode-session-end.sh` was fixed **by hand** in a past PR, and the guard was widened to all of `palinode/` at the same time — but not to `examples/`. Nothing stops the same mistake landing there again.

**Two things stand in the way, and the second is where the work is.**

1. `_SOURCE_ROOTS` is `("palinode", "tests", "bench")`, and `test_no_issue_refs_in_package_strings` walks `palinode/` alone. `examples/` is in neither.
2. **The `_SOURCE_ROOTS` scans cover `*.py` only** — `rglob("*.py")`, read with `tokenize` for comments and `ast` for docstrings and string constants. `examples/` contains **no `.py` at all**; it is `.sh`, `.md`, `.json` and `.txt`. So adding it to `_SOURCE_ROOTS` would scan nothing.

**The good news: the module already does what you need, twice.** `test_no_issue_refs_in_root_config` reads `.gitattributes` and `.gitignore` line-by-line as plain text, and `test_no_issue_refs_in_ci_workflows` does the same for `.github/workflows/*.yml`. **This is an extension of an existing pattern, not a new capability** — follow those two.

**To be clear about what this is:** `grep` finds **no** bare refs in `examples/` on `main` right now. This is a missing guard, not a live bug. Your test should fail if someone *reintroduces* one — so the way to check your work is to add a bare ref to a file in `examples/` locally, watch it fail, then remove it.

**What is wanted.**

1. A check that `examples/` carries no bare issue refs, reading files as text in the style of the two tests named above.
2. Reuse `_issue_refs` and the module's existing patterns rather than writing new regexes. The full-URL form is the intended escape hatch and must keep passing; the bare `#NNNN` form must fail. That distinction is already encoded.
3. **One judgement is yours: whether `.md` in `examples/` is in scope.** There is a precedent to reason from rather than a coin to flip — read the comment above `_ROOT_CONFIG_FILES`. It explains that the allowlist is deliberately short because `docs/CHANGELOG.md` cites issues *on purpose* and a wider prose scrub "has not been made" as a decision. Whatever you conclude, **say which you chose and why in the PR**, and point at that reasoning.

**Out of scope:** changing the existing `palinode/` guards, and fixing any ref you find outside `examples/` — open a separate issue for that.

**Tests:** the new check, plus a negative control in the style of the module's existing ones — assert the allowed full-URL form does **not** trip it.

**Reserved for @Ts-Boom**, who asked for work on #223 minutes after it was claimed. This is not open for another claim while that reservation stands. If you are interested too, you may comment to be next in line, but please do not start a PR: I will offer it to someone else only after @Ts-Boom declines or I explicitly release the reservation on this issue.

Ts-Boom:
Hi @Paul-Kyle,

This PR implements the issue-reference guard for the `examples/` directory as requested in #224.

**Key changes & implementation details:**
- Extended the issue-ref guards to scan all non-Python collateral under `examples/` (covering `.sh`, `.json`, `.txt`, and `.md` files) line by line, mirroring the pattern used for root configuration files.
- Reused existing regexes and patterns (`_issue_refs`, `_ISSUE_REF`, etc.) directly from sibling modules with zero redundancy.
- Included comprehensive test coverage in a new test module/extension covering happy paths (full public URLs), negative controls (bare `#NNN` numbers and foreign repo URLs), edge cases (adjacent URLs and plain shell hashes), and a sanity check to ensure the directory is actively scanned.

Fixes #224


Paul-Kyle:
Reviewed — thank you for extending the existing text-scan pattern and carrying its URL distinctions through to `examples/`. `test_examples_dir_is_scanned` is a useful extra check: it catches a missing or renamed directory.

I found two guard details to close before merge:

1. `_scan_file` catches `UnicodeDecodeError` and `OSError` and returns `[]`. If one selected file cannot be read, the guard silently treats that file as clean; the directory sanity test still passes because other files exist. Please fail closed instead — let the read error fail the test, or raise an assertion naming the path — and add a focused regression test using an invalid-UTF-8 selected file.
2. The extension comment says future `examples/*.py` files are already caught by the source-comment and string-constant guards. They are not: `_SOURCE_ROOTS` excludes `examples/`. Please either include `.py` in this text scan or correct the comment and state that boundary explicitly. I lean toward including it so the guard continues to cover example source files if one is added.

Your Markdown scope decision is reasoned and defensible. One non-blocking thought for later: `examples/sample-memory/` is sample store prose, and a real memory may cite an issue; if a future sample needs that, revisit the scope then. No changelog entry is needed from you for this test-only change.

Thanks for the careful coverage. Once those two details are addressed, I will recheck the updated head.

PR #227, opened 2026-09-19
docs: document search threshold provenance

### LINKED ISSUE #223: docs: say that the search thresholds were measured on bge-m3
The README says all models are swappable, and for embeddings that is true of the wiring. What it does not say is that two default search floors were measured on one model, while trigger similarity has a separate model-sensitive floor.

`search.mcp_threshold` (0.4) and `search.api_threshold` (0.5) are absolute cosine-similarity floors. The `SearchConfig` docstring in `palinode/core/config.py` records that both were measured against real `bge-m3` embeddings. Triggers also have a default cosine-similarity threshold (0.75), but the repo does not record a calibration for that default. Cosine similarities from different embedding models are not distributed the same way, so a floor that keeps 98% of true matches on one model can keep far fewer — or let far more through — on another. **We have not measured any other model, so this is not a claim that the defaults are wrong elsewhere — only that nobody has checked, and the docs should say so.**

Right now that provenance lives only in a source docstring. Someone following the README to swap models has no way to know the thresholds are worth a second look.

**What is wanted — documentation only, no code:**

1. In `README.md`, next to "All models are swappable", one or two sentences: the two default search floors were measured on `bge-m3`; if you change the embedding model, re-check them and review any trigger threshold too. Do not imply the 0.75 trigger default was measured on `bge-m3`.
2. In `palinode.config.yaml.example`, on the `search:` block's `mcp_threshold` / `api_threshold` comments, note the `bge-m3` provenance of those two search floors in a line.
3. Point at how to re-check search floors: `python -m bench.abstention` runs against whatever embedding endpoint is configured and sweeps the search floor, but it does not calibrate trigger thresholds. Read its module docstring first and describe it accurately — what it measures and what it does not.

**Related:** #222 covers the defect of switching models on an existing index. This issue is about what is still true after that is handled — a clean index in a new model, with floors tuned for the old one.

**Out of scope:** changing any default, re-measuring anything, adding per-model presets.

TomTang701:
Closes #223

## Summary

- Document that `search.mcp_threshold=0.4` and `search.api_threshold=0.5` were measured against real `bge-m3` embeddings.
- Explain that changing the embedding model should trigger a re-check of those search floors and a separate review of trigger thresholds.
- Point users to `python -m bench.abstention` and describe its measured scope from the module docstring: no-answer false positives and answer-present control retention across per-arm search thresholds.
- Add the same `bge-m3` provenance note to `palinode.config.yaml.example`.

The abstention description follows `bench/abstention.py`'s module docstring. No discrepancy with the issue summary was found. This PR changes documentation only; runtime behavior and defaults are unchanged.

## Verification

- `git diff --check`
- YAML parsing confirmed the example config and preserved `mcp_threshold: 0.4` and `api_threshold: 0.5`.
- The commit contains only `README.md` and `palinode.config.yaml.example`.
- The focused pytest could not be collected in the local environment because `httpx` is not installed; no test assertions ran.
- The live abstention benchmark was not run because this issue is documentation-only and the benchmark requires a configured embedding endpoint.


Paul-Kyle:
Reviewed — this is accurate and it does the hard part right.

The distinction you kept is the one that matters: the calibration is attributed to the two search floors, and the trigger default is named as *outside* it rather than swept in. That is exactly where this issue was easiest to get wrong.

Your `bench.abstention` description also matches the module docstring rather than my summary of it — including "per-arm", which is the docstring's own word. That is what I was asking for when I said to read it first.

No changes requested. No changelog entry needed from you: public contributions get their bullet when the change is ported, so that is on me, not on this PR.

Paul-Kyle:
Merged — thank you. The part worth naming: you kept the trigger default outside the `bge-m3` calibration instead of sweeping it in, and you described `bench.abstention` from its own docstring rather than from my summary of it. That is the failure mode this issue was filed to fix, and you avoided it without being told twice.

